PR #9, opened 2020-12-07
Feature/dc 694 simplify base app logic

oleksii-symon-corva-ai:
[DC-694](https://corvaqa.atlassian.net/browse/DC-694)

Main change: reworked and simplified base app logic.

- App changes:
-- added Context object to allow custom parameter sets for different app subclasses
-- added EventLoader for each type of Event for simplicity


corva/app/context.py
```diff
@@ -0,0 +1,19 @@
+from pydantic import BaseModel
```
finomayato, line 1: can we add a docstring. What is it `context` in the scope of this project and why do we need it?

corva/event/event.py
```diff
@@ -0,0 +1,21 @@
...
+from typing import List
+
+from corva.event.data.base import BaseEventData
+
+
+class Event(UserList):
+    def __init__(self, data: List[BaseEventData]):
+        super().__init__(data)
+
+    def __eq__(self, other):
```
finomayato, line 13: why do we need to implement `__eq__` method?

corva/event/loader/base.py
```diff
@@ -0,0 +1,28 @@
...
+    def load(self, event: str) -> Event:
+        pass
+
+    @staticmethod
+    def _load_json(event: str) -> EVENT_TYPE:
+        if not event:
+            raise ValueError('Empty event')
+
+        if not isinstance(event, str):
+            raise TypeError(f'Unknown event type {type(event)}')
```
finomayato, line 21: `json.loads` already throws an error if it cannot convert a string to json. Let's not double-check it

tests/app/test_base.py
```diff
@@ -1,262 +1,164 @@
...
 from corva.event.data.base import BaseEventData
-from corva.utils import get_state_key
+from corva.event.event import Event
+from tests.conftest import ComparableException, APP_KEY, CACHE_URL
+
+
+@pytest.fixture(scope='function')
+def base_app(mocker: MockerFixture):
+    # patch __abstractmethods__, so we can initialize BaseApp
+    mocker.patch.object(BaseApp, '__abstractmethods__', set())
```
finomayato, line 13: worth a better descriptionof why do we patch abstractmethods?

tests/app/test_base.py
```diff
@@ -1,262 +1,164 @@
...
 
+    assert (
+         getattr(BaseApp, '__abstractmethods__')
+         ==
+         frozenset([
+             'event_loader',
+             'group_by_field',
+             'get_context'
+         ])
+    )
```
finomayato, line 30: why do we need a check of which methods are abstract?

tests/app/test_base.py
```diff
@@ -1,262 +1,164 @@
...
-    state = 'state'
-
-    with patch.object(base_app, 'pre_process'), \
-         patch.object(base_app, 'process'), \
-         patch.object(base_app, 'post_process', side_effect=Exception), \
-         patch.object(base_app, 'logger') as logger_mock, \
-         pytest.raises(Exception):
-        base_app._run(event=event, state=state)
-    logger_mock.error.assert_called_once_with('An error occurred in post_process.')
+def test__run_exc_in_process(mocker: MockerFixture, base_app):
```
finomayato, line 112: why there are two underscores

finomayato: also, what this test is testing. can you please add a description or change a name, so it would be more clear?

finomayato (commented):
It's better to avoid testing private methods
https://stackoverflow.com/questions/27701183/should-i-unittest-private-protected-method

corva/app/context.py
oleksii-symon-corva-ai: Q1: What is the purpose of the python-sdk?
A1: Help companies develop custom apps easier.

Q2: Why can't Corva develop custom apps for companies?
A2: Companies have secret proprietary algorithms that they don't want to expose. The best Corva can do is to provide a tool for companies to easily write their custom apps.

Q3: What types of custom apps do companies want?
A3: Stream, Scheduled and Task apps.

Q4: What is BaseApp class in python-sdk?
A4: Each type of app executes three main steps: pre_process, process and post_process. BaseApp provides an basic interface, that each subclass (e.g TaskApp) is able to extend.

Q5: What was the problem that Context class solved?
A5: Each child class of BaseApp needed unique set of parameters to run (e.g. StreamApp needed state param, whereas TaskApp needed task_data param). Also TaskApp needed a way to pass additional data from `process` to `post_process`.

Problem:
```python
class BaseApp:
    def run(self):
        event = self.get_event()
        state = self.get_state()  # how to initialize for TaskApp that doesn't need this
        task_data = self.get_task_data()  # how to initialize for StreamApp that doesn't need this?

        self.pre_process(event=event, state=state, task_data=task_data)
        self.process(event=event, state=state, task_data=task_data)
        self.post_process(event=event, state=state, task_data=task_data)
    
    # bloated function calls below
    def pre_process(self, event, state, task_data): ...
    def process(self, event, state, task_data): ...
    def post_process(self, event, state, task_data): ...

class StreamApp(BaseApp):
    # no need in task_data param in below functions
    def pre_process(self, event, state, **kwargs): ...
    def process(self, event, state, **kwargs): ...
    def post_process(self, event, state, **kwargs): ...

class TaskApp(BaseApp):
    # no need in state param in below functions
    def pre_process(self, event, task_data, **kwargs): ...
    def process(self, event, task_data, **kwargs): ...
    def post_process(self, event, task_data, **kwargs): ...
```
Solution:
```python
class BaseContext:
    event: 'Event'

class StreamContext(BaseContext):
    state: 'State'

class TaskContext(BaseContext):
    task_data: 'TaskData'
    process_result: dict = {}  # will be set in process function

class BaseApp:
    def run(self):
        context = self.get_context()
        
        # clean function calls below
        self.pre_process(context)
        self.process(context)
        self.post_process(context)

    def get_context(self) -> BaseContext:
        """Returns all params needed for app to run in a BaseContext object"""
        ...

    def pre_process(self, context: BaseContext): ...
    def process(self, context: BaseContext): ...
    def post_process(self, context: BaseContext): ...

class StreamApp(BaseApp):
    def get_context(self) -> BaseContext: ... # returns StreamContext, knows how to initialize it

    # below functions only get parameters that they actually need
    def pre_process(self, context: BaseContext): ...
    def process(self, context: BaseContext): ...
    def post_process(self, context: BaseContext): ...

class TaskApp(BaseApp):
    def get_context(self) -> BaseContext: ...  # returns TaskContext, knows how to initialize it

    # below functions only get parameters that they actually need
    def pre_process(self, context: BaseContext): ...
    def process(self, context: BaseContext): ...
    def post_process(self, context: BaseContext): ...
```
In the same way companies will be able to write their own apps, with custom parameters passed to process function using Context.

corva/event/event.py
oleksii-symon-corva-ai: Actually, it is not needed, will delete it

tests/app/test_base.py
oleksii-symon-corva-ai: We test them to verify, that we didn't forget to make selected methods abstract and that they must be overridden in child  classes

tests/app/test_base.py
oleksii-symon-corva-ai: Decided to go with two, because the method tested is called `_run` and there is also a method `run` in same class

tests/app/test_base.py
finomayato: We don't need to verify the structure and signature of the class or its methods. It's static.
Generally, when we write the test when we need to test:
* logic
* interface (parts of the application which will be exposed to the user or to other components of the system)
And we do not need to test:
* ourselves (how do we write code and what methods we use)
* implementation details (how a programming language works, how libraries works)

Here is an example of the possible scenario: one of the child classes doesn't really need this method to be implemented and we remove `abstractmethod` and use `return NotImplemented` instead. Logically, nothing should break. Though, this test will fail.

corva/app/context.py
oleksii-symon-corva-ai: Also added a docstring to BaseContext

tests/app/test_base.py
oleksii-symon-corva-ai: Expanded the comment

corva/app/context.py
finomayato: perfect! Thank you!

The one comment: let's keep context near the class itself. We don't really need another module to keep all contexts

oleksii-symon-corva-ai: I've moved these models to appropriate corva/models/ dirs in [pull 7](https://github.com/corva-ai/python-sdk/pull/7)

tests/app/test_base.py
oleksii-symon-corva-ai: Thanks, deleted the test

corva/app/stream.py
```diff
@@ -1,100 +1,96 @@
...
             [
                 record.measured_depth
                 for record in all_records
                 if record.measured_depth is not None
             ],
-            default=self.DEFAULT_LAST_PROCESSED_DEPTH
+            default=self.DEFAULT_LAST_PROCESSED_VALUE
         )
 
         mapping = {'last_processed_timestamp': last_processed_timestamp,
```
ville, line 82: Not directly related to this PR but no need to store to cache if the value is missing. Depth based data never has timestamp (and vice versa).

ville (approved):

PR #10, opened 2020-12-08
DC-673 Pypi build and deploy

jordanambra:
### Summary
This PR is for adding general build and deployment functionality for the Corva Python SDK so that it is automatically tested and deployed as we provide updates.

### Tools Involved
* flake8 for static analysis
* `pytest` for automated testing
* `coverage` for code coverage reporting
* GitHub Actions as the build runner
* Deploy to PyPi


.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
+name: CI
+
+on: push
```
finomayato, line 3: do we really want to execute workflows on every push to every branch? it makes sense for lint and tests, but not for  deployment job

requirements.txt
```diff
@@ -1,3 +1,4 @@
+coverage==5.3
```
finomayato, line 1: We should have separate `requirements.txt` and `test_requirements.txt` for the production and testing environment accordingly

setup.py
```diff
@@ -0,0 +1,43 @@
...
+    description='SDK for interacting with Corva',
+    long_description=readme,
+    long_description_content_type="text/markdown",
+    keywords='corva, sdk',
+    packages=setuptools.find_packages(exclude=[".*", "data", "test"]),
+    install_requires=[
+        "pydantic >= 1.7.2",
+        "redis >= 3.5.3",
+        "requests >= 2.25.0",
+        "urllib >= 1.26.2"
```
finomayato, line 40: we could read `requirements.txt` file here

.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Cache pip dependencies
```
finomayato, line 37: why do we need to cache pip dependencies?

finomayato (changes requested):
1. I cannot install this package locally:
image

2. Right now we have two places where we are copying requirements:
* `setup.py`
* `requirements.txt`

This could lead to potential discrepancies and issues since tests are using the requirements file but the module itself - requirements provided in setup.

Also, it's bad to have testing and production requirements in the same `requirements.txt` file. Usually, systems have `requirements.txt` and `test_requirements.txt`

setup.py
finomayato, line 40: also, we already have discrepancies between packages in the requirements file, and these ones
```suggestion
        "urllib3 ~= 1.26.2"
```

.github/workflows/main.yml
jordanambra: For subsequent build performance

.github/workflows/main.yml
jordanambra: The deployment job has an if statement to filter it only to tags.

finomayato: it would be much safer to create another workflow dedicated to deployment job and not rely on the conditional expression

requirements.txt
jordanambra: The production or installed environment uses the setup.py `install_requires` field. In public Python packages, there are different ways of doing this, but there's nothing wrong with this particular choice. If you are installing from the requirements.txt, it's because you are trying to develop locally, so you would need everything.

setup.py
jordanambra: I'd rather not combine these. Sometimes the requirements.txt for development will require different versions and even different version locking than `install_requires`, which may have looser constraints to play nicely with other required libraries for a project.

For example, in requirements.txt we lock to a certain version, but in install_requires we just want a minimum version.

.github/workflows/main.yml
finomayato: This step complicates the understanding of how are we building the project while we don't have any reason to reduce build time. I'd rather not use it. Besides, it's cache - always tricky )

jordanambra: Can you explain the potential complications? This is common practice both in terms of CI and in terms of GitHub Actions, both of which recommend caching dependencies, and it's pretty much automatically handled.

setup.py
jordanambra: Fixed, thank you

jordanambra:
#1 - Thanks, this is fixed now
#2 - I explained this in a separate comment, but since requirements.txt and setup.py serve slightly different purposes, at least in the future, it's OK that they are similar, yet copied.
#3 - That's true in a first-party project, but libraries are different because setup.py performs the same function that the normal separation of requirements would do.

setup.py
finomayato: > Sometimes the requirements.txt for development will require different versions and even different version locking than install_requires

Why? It does not make sense to me

In that case, we will face issues with requirements discrepancies we just fixed. And the worst part is that tests won't catch this error, since tests are using development requirements. We will be required to catch these issues when our lib already installable via pip for users. In the worst-case scenario, a minor version library upgrade will break the user's application

.github/workflows/main.yml
jordanambra: I don't think GitHub Actions supports workflow interdependencies, where one workflow triggers another once completed, based on the branch.

If we make another workflow dedicated to the deployment job, it would be 80% copied code from this.

This is pretty standard practice though.

finomayato:
> 2. I explained this in a separate comment, but since requirements.txt and setup.py serve slightly different purposes, at least in the future, it's OK that they are similar, yet copied.

They both contain required packages for successful program execution. The one we use to develop/test and another to pack dependencies in a package.
Complications of having 2 (not programmatically tied together) places to introduce requirements for the application:
1. Developer always need to remember to add requirements into 2 places
2. Development environment have more discrepancies with the production environment
3. Tests are using the `requirements.txt` file. This means that tests could pass successfully, but the packaged library will fail on execution

> 3. That's true in a first-party project, but libraries are different because setup.py performs the same function that the normal separation of requirements would do.

You right about different approaches for libraries. But there are reasons to go with a solution different from the current. **The proposed solution** is to divide requirements into `requirements.txt` (corresponding production) and `test_requirements.txt` (test requirements). Use former in `setup.py`. In that way, we will minimize discrepancies between dev, test, and production and make sure the issue we just fixed never happens again. If you do not like this solution - we can look for other possible solutions. But for now, it looks like the easiest one.

setup.py
jordanambra: We can't version lock somebody to an exact version like we can (and should) with requirements.txt, since that may prevent them from installing. They serve slightly different purposes.

Eventually we can use something like a tox matrix that would flush out those sorts of errors. This is a first step.

.github/workflows/main.yml
finomayato: The complication is: we have a piece of code we don't need. And while debugging deployment issues - it will be another step to read and check. 

At the same time, I have no strong preferences over this one. We can go with cache if you'd like

setup.py
finomayato: why do we need to lock something to an exact version in the development, and do not do that in production?

.github/workflows/main.yml
finomayato: makes sense

setup.py
jordanambra: Let's say the `requests` library breaks something in version 1.2.3 that doesn't get fixed until 1.3.0. If we version lock to exactly 1.2.3, or 1.2.4, or whatever, then they will be unable to version lock `requests` in their own requirements.txt and have it be compatible with the `corva` package, and pip install will fail.

If we instead use `requests >= 1.2.0` or something similar, then we allow some flexibility.

However, in our own requirements.txt, we want to lock so that we have deterministic test consistency and a known-good version.

finomayato: 1. We should limit our versions range only to minor versions. Judging by [semver](https://semver.org/) convention, major versions could change their user interface, which will break our library.

2. My proposal includes using version ranges in development too. I do not see any downsides. It's closer to how users would use it. Let's worry about users, not about tests. The sole purpose of tests is to catch issues before releasing the library. If they don't - who cares if they are consistent and working with known-good versions? Do you agree with me?

jordanambra: The point of the requirements.txt being pinned is to provide users and tests with a known working version. If you don't know exactly which version you are building and testing on CI with, then you haven't proven anything with your tests. So I agree 100% with your philosophy of point 2, and am trying to ensure we get there.

This method is common best practice for developing a library for public consumption, do we have a special case that would require us to reinvent the wheel?

Flask: https://github.com/pallets/flask/blob/master/setup.py
DRF: https://github.com/encode/django-rest-framework/blob/master/setup.py
httpx: https://github.com/encode/httpx/blob/master/setup.py
requests: https://github.com/psf/requests/blob/master/setup.py
Scrapy: https://github.com/scrapy/scrapy/blob/master/setup.py
Pandas: https://github.com/pandas-dev/pandas/blob/master/setup.py

I think we should move towards pyproject.toml and Poetry, or something similar, but it's a bit too new IMO.

finomayato: > The point of the requirements.txt being pinned is to provide users and tests with a known working version

Correct. 
In order to provide a user with known working versions - we should have the same version ranges/pins in tests. We can say that version is working when we tested it. Isn't that right?

jordanambra: No, and that's where there's a discrepancy. We can do that for setup.py, but not in the requirements.txt at the time of tests, because currently we are not testing every single combination of 3rd party libraries, nor do we want to for tests. Therefore, we must version lock to a specific pinned version.

We can test a fixed set of 3rd party libraries, at a given point in time, and stamp a release of ours as guaranteed working with that combo, but we want to provide flexibility to our actual users, who may have a different version that still works, but that we aren't going to test.

Who are we to tell users that they can't pin a version of a library we also share, as long as it works for everyone?

This is by far the most common way to handle setup.py and an accepted best practice for 3rd party libraries on PyPI. I don't think this deserves any more time, especially for something that we are intentionally keeping separate.

finomayato: well, I cannot agree with you. I don't see the logic in permitting users to install versions of packages our lib is using that are not compatible. If they want to - they will see the warning from `pip`. But if I'm the user and I install our package which is not working because of requirements... well, that is not good

oleksii-symon-corva-ai (approved):

ville (approved):

jordanambra:
I understand there are some outstanding concerns and opinions about how setup.py and requirements.txt should be handled. For now, these are not dealbreakers for either our workflow or our deployment process, so they should be converted into separate discussions, if we wish to address them at all.

Since this task is a blocker for the release of our product, and we can certainly develop, test, and deploy with it how it is, I'm going to merge.

Thank you everyone for your reviews.

PR #11, opened 2020-12-23
Concept for simplified UI

oleksii-symon-corva-ai:
[DC-751](https://corvaqa.atlassian.net/browse/DC-751)
- added UI usage examples

### DIFF
--- a/docs_src/tutorial_1_hello_world.py
+++ b/docs_src/tutorial_1_hello_world.py
@@ -17,4 +17,4 @@ def lambda_handler(event, context):
 
     """AWS lambda handler"""
 
-    user_job.run(event)  # 5 run your function
+    user_job(event)  # 5 run your function
--- a/docs_src/tutorial_1_hello_world.py
+++ b/docs_src/tutorial_1_hello_world.py
@@ -17,4 +17,4 @@ def lambda_handler(event, context):
 
     """AWS lambda handler"""
 
-    user_job(event)  # 5 run your function
+    user_job(event)  # 5 pass only event as parameter to your function call
--- a/docs_src/tutorial_1_hello_world.py
+++ b/docs_src/tutorial_1_hello_world.py
@@ -1,11 +1,11 @@
-from corva import Api, Corva, Event, State
+from corva import Api, Corva, StreamEvent, State
 
 app = Corva()  # 1 initialize the app
 
 
 @app.stream  # 2 add decorator with needed event type to your function
-def user_job(event: Event, api: Api, state: State):
-    # 3 add parameters with predefined types, that will be injected automatically
+def stream_app(event: StreamEvent, api: Api, state: State):
+    # 3 above, add parameters with predefined types, that will be injected automatically
 
     """User's main logic function"""
 
@@ -17,4 +17,4 @@ def lambda_handler(event, context):
 
     """AWS lambda handler"""
 
-    user_job(event)  # 5 pass only event as parameter to your function call
+    stream_app(event)  # 5 pass only event as parameter to your function call
--- a/docs_src/tutorial_1_hello_world.py
+++ b/docs_src/tutorial_1_hello_world.py
@@ -1,0 +1,20 @@
+from corva import Api, Corva, Event, State
+
+app = Corva()  # 1 initialize the app
+
+
+@app.stream  # 2 add decorator with needed event type to your function
+def user_job(event: Event, api: Api, state: State):
+    # 3 add parameters with predefined types, that will be injected automatically
+
+    """User's main logic function"""
+
+    pass
+
+
+def lambda_handler(event, context):
+    # 4 define function that will be run by AWS lambda
+
+    """AWS lambda handler"""
+
+    user_job.run(event)  # 5 run your function
--- a/docs_src/tutorial_2_configuration.py
+++ b/docs_src/tutorial_2_configuration.py
@@ -1,4 +1,4 @@
-from corva import Api, Corva, Event, State
+from corva import Api, Corva, StreamEvent, State
 
 app = Corva(
     # 1 api params
@@ -14,7 +14,7 @@ app = Corva(
 
 
 @app.stream
-def user_job(event: Event, api: Api, state: State):
+def stream_app(event: StreamEvent, api: Api, state: State):
     """User's main logic function"""
 
     pass
@@ -23,4 +23,4 @@ def user_job(event: Event, api: Api, state: State):
 def lambda_handler(event, context):
     """AWS lambda handler"""
 
-    user_job(event)
+    stream_app(event)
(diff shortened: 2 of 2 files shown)


docs_src/tutorial_2_configuration.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # 3 state params
+    state_url = 'redis://'
+    state_params = {'param1': 'val1'}
+
+    # 4 other params
+    ...
+
+
+app.config_from_object(Config)
```
finomayato, line 27: we do not need a custom class for the configuration, we can pass them to the `Corva` initialization:
```
app = Corva(api_url='api.localhost', state_url='redis://', ...)
```
It would be simpler to start from this point

docs_src/tutorial_1_hello_world.py
```diff
@@ -0,0 +1,20 @@
...
+
+    pass
+
+
+def lambda_handler(event, context):
+    # 4 define function that will be run by AWS lambda
+
+    """AWS lambda handler"""
+
+    user_job.run(event)  # 5 run your function
```
finomayato, line 20: still, I think it is a bad design to expose information about internal infrastructure to the user

finomayato:
Overall looks good, but there are some moments that I do not recommend:
* Exposing information about internal infrastructure
* SDK configuration via the class (easier to propagate configurations right into the `__init__`)

docs_src/tutorial_1_hello_world.py
oleksii-symon-corva-ai: do you mean `lambda_handler` or `user_job.run`?

finomayato: I mean `lambda_handler`. It exposes information about lambda usage. Everything is ok with `user_job.run` - it's just part of the user-interface and has no redundant information for a user

finomayato: Well, almost everything is ok. `user_job` is a function. It has no methods. So, this line won't work

oleksii-symon-corva-ai: will make it work like this [link](https://docs.celeryproject.org/en/stable/getting-started/first-steps-with-celery.html#calling-the-task)

oleksii-symon-corva-ai: I agree about `lambda_handler`, but it has to do to more with how dev center works, not sdk

finomayato: > I agree about lambda_handler, but it has to do to more with how dev center works, not sdk

It is possible to change how the dev center works

> will make it work like this link

why we cannot just call a function as it is? `user_job(event)`

oleksii-symon-corva-ai: > It is possible to change how the dev center works. How hard it will be to change it on the dev center side?

oleksii-symon-corva-ai: > It is possible to change how the dev center works. How hard it will be to change it on the dev center side?

@jordanambra @ville 

> why we cannot just call a function as it is? user_job(event)

we can

ville:
We will let the Ccustomers know that the apps run in Lambda because that sets some restrictions on how the apps must be built. They also need to know what's in the context and they must be able to use serverless to test out their code.

finomayato (approved):
@ville thanks for letting me know what is the high-level plan

Approved, assuming that we sacrifice the UX of our SDK to validate SDK idea faster

docs_src/tutorial_1_hello_world.py
```diff
@@ -0,0 +1,20 @@
...
+
+    pass
+
+
+def lambda_handler(event, context):
+    # 4 define function that will be run by AWS lambda
+
+    """AWS lambda handler"""
+
+    user_job(event)  # 5 run your function
```
jordanambra, line 20: It's confusing that in the function invocation, you are only passing the event, but there is no mention of DI for the other parameters. I think that should be documented, or dealt with differently.

oleksii-symon-corva-ai: Yeah, I will document this

oleksii-symon-corva-ai: What is DI?

docs_src/tutorial_2_configuration.py
```diff
@@ -0,0 +1,26 @@
+from corva import Api, Corva, Event, State
+
+app = Corva(
+    # 1 api params
+    api_url='api.localhost',
```
ville, line 5: Are these params pulled from env variables by default?

oleksii-symon-corva-ai: yes, they will be provided by default, with possibility to override

also some unique settings for each app type will be moved to `stream`, `task` and `scheduled` decorators

docs_src/tutorial_1_hello_world.py
```diff
@@ -0,0 +1,20 @@
+from corva import Api, Corva, Event, State
```
jordanambra, line 1: from corva import Corva is a little odd. Can we rename that container?

docs_src/tutorial_1_hello_world.py
```diff
@@ -0,0 +1,20 @@
...
+
+app = Corva()  # 1 initialize the app
+
+
+@app.stream  # 2 add decorator with needed event type to your function
+def user_job(event: Event, api: Api, state: State):
+    # 3 add parameters with predefined types, that will be injected automatically
```
jordanambra, line 8: This line isn't clear that it's referring to the above parameters

docs_src/tutorial_1_hello_world.py
```diff
@@ -0,0 +1,20 @@
+from corva import Api, Corva, Event, State
+
+app = Corva()  # 1 initialize the app
+
+
+@app.stream  # 2 add decorator with needed event type to your function
+def user_job(event: Event, api: Api, state: State):
```
jordanambra, line 7: Something to keep in mind is that people will probably be copying this code, or using one of our templates that will look like this, so a more useful name that `user_job` would be good.

jordanambra (changes requested):

docs_src/tutorial_1_hello_world.py
oleksii-symon-corva-ai: maybe we can leave it as it is. other popular libraries use the same approach e.g.
```python3
from celery import Celery
from fastapi import FastAPI
```
if not, what would be a better naming?

docs_src/tutorial_1_hello_world.py
oleksii-symon-corva-ai: renamed to `stream_app`

docs_src/tutorial_1_hello_world.py
oleksii-symon-corva-ai: expanded the comment

PR #12, opened 2020-12-24
Feature/dc 752 develop corva class

oleksii-symon-corva-ai:
[DC-752](https://corvaqa.atlassian.net/browse/DC-752)

- Added Corva class
- Added some middlewares
- Deleted useless files and dirs


corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+        call = wrapper_factory(mw, call)
+
+    return call
+
+
+class Corva:
+    def __init__(
+         self,
+         *,
+         middleware: Optional[List[MIDDLEWARE_TYPE[BaseContext]]] = None
```
finomayato, line 30:
```suggestion
         middlewares: Optional[List[MIDDLEWARE_TYPE[BaseContext]]] = None
```

corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+     call: Callable,
+     middleware: Optional[List[MIDDLEWARE_TYPE]] = None
+) -> MIDDLEWARE_CALL_TYPE:
+    def wrapper_factory(mw, call):
+        def wrapper(ctx):
+            return mw(ctx, call)
+
+        return wrapper
+
+    middleware = middleware or []
```
finomayato, line 18: should be plural

corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+
+from corva.middleware.unpack_context import unpack_context
+from corva.models.base import BaseContext
+from corva.types import MIDDLEWARE_CALL_TYPE, MIDDLEWARE_TYPE
+
+
+def wrap_call_in_middleware(
+     call: Callable,
+     middleware: Optional[List[MIDDLEWARE_TYPE]] = None
+) -> MIDDLEWARE_CALL_TYPE:
```
finomayato, line 11: what is the difference between `Callable` and `MIDDLEWARE_CALL_TYPE` types?

corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+    def get_middleware_stack(
+         self,
+         middleware: Optional[List[MIDDLEWARE_TYPE[BaseContext]]] = None
+    ) -> List[MIDDLEWARE_TYPE[BaseContext]]:
+        middleware = middleware or []
+
+        middleware_stack = (
+             middleware
+             + self.user_middleware
+             + [unpack_context]
```
finomayato, line 43: what is `unpack_context`? it makes no sense in the scope of this method and middleware stack

corva/middleware/unpack_context.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Callable, Optional, Tuple
+
+from pydantic.utils import lenient_issubclass
+
+from corva.models.base import BaseContext, BaseEvent
+from corva.network.api import Api
+from corva.state.redis_state import RedisState
+
+
+def _parse_call(call: Callable) -> Tuple[Optional[str], Optional[str], Optional[str], Optional[str]]:
```
finomayato, line 11: what is the purpose of this function?

corva/middleware/unpack_context.py
```diff
@@ -0,0 +1,49 @@
+import inspect
```
finomayato, line 1: I do not understand what this whole module is for

corva/types.py
```diff
@@ -1,5 +1,6 @@
-from typing import List, Union
+from typing import Any, Callable, List, TypeVar, Union
 
+_T = TypeVar('_T')
```
finomayato, line 3: what is this? why do we need it?

tests/app/test_scheduled.py
```diff
@@ -61,6 +61,7 @@ def test_group_by_field():
     assert ScheduledApp.group_by_field == 'app_connection_id'
 
 
+@pytest.mark.skip(reason='No need to run this as new architecture is being developed.')
```
finomayato, line 64: let's delete it then?

corva/application.py
oleksii-symon-corva-ai: it is used in this form everywhere I saw. maybe because of [this](https://english.stackexchange.com/a/257139)

corva/application.py
oleksii-symon-corva-ai: `Callable` doesn't specify arguments, that callable accepts. whereas `MIDDLEWARE_CALL_TYPE` provides them

corva/application.py
oleksii-symon-corva-ai: it is a default middleware, that should always be called last. it scans parameter annotations of user function and puts values from context into argument names which have needed annotations. [link](https://github.com/corva-ai/python-sdk/blob/7e8a5aa1ceaf37e7447693b05cbc8f6c4ae0459a/docs_src/tutorial_1_hello_world.py#L7)

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: it scans parameter annotations of user function and finds names of parameters which have needed annotations. allows this functionality [link](https://github.com/corva-ai/python-sdk/blob/7e8a5aa1ceaf37e7447693b05cbc8f6c4ae0459a/docs_src/tutorial_1_hello_world.py#L7)

oleksii-symon-corva-ai: inspired by [this](https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L336)

corva/application.py
oleksii-symon-corva-ai: inspired by [this](https://github.com/tiangolo/fastapi/blob/master/fastapi/dependencies/utils.py#L336)

tests/app/test_scheduled.py
oleksii-symon-corva-ai: I dont want to delete them just yet, because code may be reused for new tests

corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(
+         self,
+         *,
+         middleware: Optional[List[MIDDLEWARE_TYPE[BaseContext]]] = None
+    ):
+        self.user_middleware = middleware or []
+
+    def get_middleware_stack(
+         self,
+         middleware: Optional[List[MIDDLEWARE_TYPE[BaseContext]]] = None
```
jordanambra, line 36: I don't think this should be Optional/None. It can default to an empty list.

corva/middleware/unpack_context.py
```diff
@@ -0,0 +1,49 @@
...
+    if event_param_name:
+        kwargs[event_param_name] = context.event
+    if state_param_name:
+        kwargs[state_param_name] = context.state
+    if api_param_name:
+        kwargs[api_param_name] = context.api
+    if context_param_name:
+        kwargs[context_param_name] = context
+
+    context.user_result = call_next(**kwargs)
```
jordanambra, line 47: What's this used for?

corva/application.py
```diff
@@ -0,0 +1,52 @@
...
+    for mw in reversed(middleware):
+        call = wrapper_factory(mw, call)
+
+    return call
+
+
+class Corva:
+    def __init__(
+         self,
+         *,
```
jordanambra, line 29: We're not doing anything with the * here? Can we do *args at least?

corva/middleware/unpack_context.py
jordanambra: This module is magical, and therefore needs a lot of comments to explain concepts, choices we made, and reasoning. Please include as much detail as you can to clarify.

jordanambra (changes requested):
I'm confused about a few things here.

1. I noticed in the documentation that we are trying to decorate the app. Is that done somewhere in this PR, or does it already exist?
2. This is combining automatic DI and middleware. If at all possible, I would like to separate those two concepts.
3. Normally with a middleware context, you can extend it with custom properties to pass around. Can you do that?

corva/middleware/unpack_context.py
finomayato: we don't need to parse it. We already know what is passed into this function

corva/application.py
finomayato: let's call it `default_middlewares` then?

tests/app/test_scheduled.py
finomayato: We have a version control system if we want to revisit older versions of a codebase

corva/application.py
oleksii-symon-corva-ai: https://github.com/corva-ai/python-sdk/pull/12#discussion_r549382361

corva/application.py
oleksii-symon-corva-ai: maybe adding comment will be enough, as `default_middleware` may be either before or after `user_middleware`. e.g [link](https://github.com/encode/starlette/blob/e4307065ea6dbe708fba9643d14fe7adfff06d46/starlette/applications.py#L79).

```suggestion
             + [unpack_context]  # default middleware, should always be called last
```

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: I'll give some examples, why unpack_context mware is needed

1. different app types need different parameter sets. e.g task apps don't need `State` param, because they are only called once

```python3
@app.task
def user_job(event: Event, api: Api): pass
```

2. parameters may be shuffled e.g

```python3
@app.stream
def user_job(api: Api, state: State, event: Event): pass
``` 

3. I also wanted to add support to pass additional `Context` param. which may not be always needed

```python3
@app.stream
def user_job(api: Api, state: State, event: Event, context: Context): pass
```

corva/application.py
oleksii-symon-corva-ai: it is not good to use mutable types as default arguments

we can do the following as alternative:

```suggestion
         middleware: typing.Sequence[MIDDLEWARE_TYPE[BaseContext]] = ()
```

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: will be used for 

1. return user the result of their function
2. automatically saving result for user in `task` app [link to current code](https://github.com/corva-ai/python-sdk/blob/4b674c24b496fd97b3aecde07fcb028388442c08/corva/app/task.py#L20)

corva/application.py
oleksii-symon-corva-ai: wanted to make arguments to Corva class keyword only, so they dont get mixed, when there are a lot of settings

oleksii-symon-corva-ai:
@jordanambra 
> I noticed in the documentation that we are trying to decorate the app. Is that done somewhere in this PR, or does it already exist?

this PR only prepares needed functionality for adding app decorators in future.

> This is combining automatic DI and middleware. If at all possible, I would like to separate those two concepts.

What is DI?

> Normally with a middleware context, you can extend it with custom properties to pass around. Can you do that?

thanks, added

corva/application.py
finomayato: why comment if we can rename the variable?

corva/application.py
finomayato: previous is better. Let's work with Optional, rather than with sequences

corva/middleware/unpack_context.py
finomayato: Different types have different logic, and that is ok
If these are positional arguments - they should not be shuffled
If these parameters are named arguments - we don't care if they are shuffled
The context could be added as an optional named parameter

From my perspective, the whole `unpack` module resolves the non-existing problem

corva/middleware/unpack_context.py
finomayato: let's call it then appropriately, "call_next" is a bad naming

corva/application.py
finomayato: That is generating more questions than helps a user. Can we go with just one keyword parameter we have for now?

finomayato (changes requested):
* Complicated functionality of unpacking parameters
* Feature of providing a possibility to add custom middlewares to the user not needed

There was a discussion about having a middleware feature for the user with @jordanambra. As a project owner, if he thinks that we need middlewares - you can disregard my second point.

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: If there is no `unpack` middleware, what do we do in the next cases?

> If these are positional arguments - they should not be shuffled

1.  arguments not shuffled but in second case state is missing

```python3
@app.stream
def user_job(event: Event, state: State, api: Api): pass

@app.stream
def user_job(event: Event, api: Api): pass
```

> If these parameters are named arguments - we don't care if they are shuffled

2. user uses different parameter name. it is probably not good to make user memorize parameter names.

```python3
@app.stream
def user_job(e: Event, s: State, a: Api): pass
```

corva/application.py
oleksii-symon-corva-ai: it is not obvious whether `default_middleware` should be placed before of after `user_middleware`, especially, since there is another `middleware` in function. we can call it `tail_middleware` though

finomayato: let's call it `default_middleware` and write a comment `user-defined middlewares should come last.`

corva/middleware/unpack_context.py
finomayato: > arguments not shuffled, but in the second case, the state is missing

They should not be missing. Why are they missing?

> user uses different parameter names. it is probably not good to make user memorize parameter names

We are sending 3 parameters into the function: `event,` `state,` and `api,` in the same order. For example, lambda does this with `event` and `context. It's lambda's responsibility to send them. But it's my responsibility, as a developer, to receive them. I can call my function whatever I want, I can call my arguments whatever I want, but the first argument will be an event, and the second one - the context. No matter how I will call them in my function. Can we do that?.`

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: Added comments

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: > They should not be missing. Why are they missing?

different app types need different parameter sets. e.g task apps don't need State param, because they are only called once

finomayato: But we have different decorators for different types. At the execution time, we know what parameters which function should receive

corva/application.py
oleksii-symon-corva-ai: fixed

corva/types.py
oleksii-symon-corva-ai: i've deleted this, and also replaced type annotations with simple Callable

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: I thought we were talking about `context.user_result`. @finomayato what would be a better name?

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: Reworked unpack_context.py to always pass predefined parameters like `event` and `context` in aws lambda functions.

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: reworked whole module

tests/app/test_scheduled.py
oleksii-symon-corva-ai: deleted skipped tests

corva/application.py
```diff
@@ -0,0 +1,40 @@
...
+
+        middleware_stack = middleware + self.user_middleware
+
+        return middleware_stack
+
+    def add_middleware(self, func: Callable) -> None:
+        self.user_middleware.append(func)
+
+    def middleware(self, func: Callable) -> None:
+        return self.add_middleware(func=func)
```
finomayato, line 40: `middleware` is redundant. `add_middleware` is more self-explanatory. Let's leave just this one

corva/middleware/unpack_context.py
```diff
@@ -0,0 +1,26 @@
+from typing import Callable
+
+from corva.models.base import BaseContext
+
+
+def unpack_context_factory(include_state=False, include_context=False):
```
finomayato, line 6: I still don't get the purpose of this function. Better to use different functions to process different tasks types

finomayato (changes requested):

corva/application.py
oleksii-symon-corva-ai: deleted middleware decorator

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: reworked the logic

corva/middleware/unpack_context.py
```diff
@@ -0,0 +1,24 @@
+from typing import Callable
+
+from corva.models.base import BaseContext
+
+
+def unpack_context_factory(include_state=False):
+    def unpack_context(context: BaseContext, call_next: Callable) -> BaseContext:
```
finomayato, line 7: can't we just move the `include_state` in the `unpack_context` function as a parameter and get rid of the nested function?

finomayato (commented):
I do not understand why do we need these functions without usage
* `unpack_context_factory`
* `wrap_call_in_middleware`

Are those a part of the user-facing interface?

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: we can, then usage will look like
```python3
unpack_context_factory(include_state=True)  # before

functools.partial(unpack_context, include_state=True)  # after
```

oleksii-symon-corva-ai:
> I do not understand why do we need these functions without usage ...

They are used in next PRs

corva/middleware/unpack_context.py
finomayato: Why do we need to have a function in the result?. Why not return value right away?

finomayato (approved):

corva/middleware/unpack_context.py
oleksii-symon-corva-ai: we need the function to wrap it into middleware layers. after wrapping middleware stack will be called and run our function at some point

corva/application.py
```diff
@@ -0,0 +1,37 @@
+from typing import Callable, List, Optional
+
+
+def wrap_call_in_middleware(
```
jordanambra, line 4: I don't think that we should copy the way FastAPI and Starlette does middleware. I know they use the call_next functionality, but that makes it so that you have interdependencies between middleware, and it's way too easy for a user to forget to properly handle the call_next value.

Can we handle this through using a list or chain, and the app itself runs all of the middleware?

Something like the pseudocode below:

```
    for middleware in middleware_list:
        context = self.run_middleware(context)

    return context
```

corva/middleware/unpack_context.py
jordanambra: Can we pass either a context or a standard interface middleware functions, and expect them to all implement the interface, instead of having to do magic DI here?

Alternately, middleware could be classes instead of functions, and you pass in all these keywords to the __init__ call, and the middleware class can then choose whether it wants to deal with them or not.

Personally I think using middleware classes is easier to understand and use, and gives more powerful functionality.

corva/models/base.py
```diff
@@ -1,27 +1,40 @@
...
+    state: Optional[RedisState] = None
+    user_result: Any = None
 
 
 class BaseEventData(BaseModel):
     class Config:
         extra = Extra.allow
+
+
+BaseEventDataTV = TypeVar('BaseEventDataTV', bound=BaseEventData)
```
jordanambra, line 34: Do we need this anymore now that this isn't nested?

jordanambra (changes requested):
I think we need to revisit the middleware part of this. I'd really like to move away from this function-based + call_next + DI approach because I think it adds a lot more maintenance complexity that I don't want to support down the road. I think it would be much easier to add middleware by making a middleware mixin or just a plain class that responds to a certain function call, like `process`, and always returns a Context.

Then, to process the middleware, we run the chain of middleware from start to finish, and then use the resulting Context in the actual app itself.

corva/application.py
oleksii-symon-corva-ai: I think deleting call_next will make life harder for us. But as for users, we can make them define special types of middleware which have only context as parameter, but then they will be able to run code that only precedes their main function.

oleksii-symon-corva-ai:
about https://github.com/corva-ai/python-sdk/pull/12#discussion_r556624367

oleksii-symon-corva-ai:
@jordanambra about https://github.com/corva-ai/python-sdk/pull/12#discussion_r556624367

I addressed this question [here](https://github.com/corva-ai/python-sdk/pull/13#discussion_r552485246)

corva/models/base.py
oleksii-symon-corva-ai: it is needed for correct type hints in ListEvent instances. ListEvent class was deleted in pull 14 at all

jordanambra (approved):

PR #13, opened 2020-12-25
Feature/dc 756 migrate loaders logic to middlewares

oleksii-symon-corva-ai:
[DC-756](https://corvaqa.atlassian.net/browse/DC-756)

- added loader middleware
- simplified event loading


corva/models/stream.py
```diff
@@ -1,27 +1,55 @@
...
-    asset_id: int
-    company_id: int
-    version: int
-    data: Dict[str, Any]
-    measured_depth: Optional[float] = None
-    collection: str
+class StreamEventData(BaseEventData):
+    class Record(BaseEventData):
+        class Data(BaseEventData):
+            hole_depth: Optional[float] = None
```
finomayato, line 14: wow, let's not do nested classes. Especially triple folding of these

corva/loader/scheduled.py
```diff
@@ -1,22 +1,15 @@
...
+    parse_as_type = List[List[ScheduledEventData]]
 
-        data = []
-        for subdata in event:
-            subdata['app_connection_id'] = subdata.pop('app_connection')
-            subdata['app_stream_id'] = subdata.pop('app_stream')
-            data.append(ScheduledEventData(**subdata))
+    def load(self, event: str) -> ScheduledEvent:
+        parsed = self.parse(event=event)  # type: ScheduledLoader.parse_as_type
+        parsed = list(chain(*parsed))
```
finomayato, line 13: what conversion are we trying to do here? what `self.parse` return to us and what do we need to have?

corva/middleware/loader.py
```diff
@@ -0,0 +1,13 @@
+from typing import Callable
+
+from corva.loader.base import BaseLoader
+from corva.models.base import BaseContext
+
+
+def loader_factory(*, loader: BaseLoader) -> Callable:
```
finomayato, line 7: why do we need this function to receive every argument before loader?

finomayato: also, this function is not used anywhere

corva/loader/base.py
```diff
@@ -1,20 +1,18 @@
...
-    @staticmethod
-    def _load_json(event: str) -> EVENT_TYPE:
-        try:
-            event = json.loads(event)
-        except ValueError as exc:
-            raise ValueError('Invalid event JSON') from exc
-
-        return event
+    @classmethod
+    def parse(cls, event: str) -> Any:
```
jordanambra, line 17: I doubt we want to return `Any` here. Maybe a union of multiple event types instead?

jordanambra: Also, I am not sure that this belongs as an inherited class. Can we have the model classes themselves each just be able to parse their own event?

corva/models/stream.py
```diff
@@ -1,27 +1,55 @@
...
+    def app_connection_id(self) -> int:
+        return self.metadata.apps[self.app_key].app_connection_id
+
+    @property
+    def app_stream_id(self) -> int:
+        return self.metadata.app_stream_id
+
+    @property
+    def is_completed(self) -> bool:
+        return self.records[-1].collection == 'wits.completed'
```
jordanambra, line 51: Wrap this with a try so we don't have an IndexError

corva/models/stream.py
jordanambra: I think these can be unpacked as their own independent classes and then set as attributes on the class they belong to.

jordanambra (changes requested):
I'm not sure about some of the class inheritance and ABC structures here. They seem a bit overly complicated for just parsing JSON into Pydantic. Can we simplify our loading schema?

corva/models/stream.py
oleksii-symon-corva-ai: unnested the classes

corva/models/stream.py
oleksii-symon-corva-ai: reworked the logic

corva/middleware/loader.py
oleksii-symon-corva-ai: deleted keyword only argument

l`oader_factory` exists to parametrize `loader_`. e.g `Stream`, `Scheduled` and `Task` apps all have different loaders, but can use same middleware

it will be used in the next pr, where I add first app type

corva/loader/scheduled.py
oleksii-symon-corva-ai: we flatten list of lists into one list. e.g flatten List[List[ScheduledEventData]] into List[ScheduledEventData]

corva/loader/base.py
oleksii-symon-corva-ai: moved loading logic to event classes

corva/loader/scheduled.py
oleksii-symon-corva-ai: also the logic is moved to corva.models.scheduled.ScheduledEvent.from_raw_event now

corva/app/scheduled.py
```diff
@@ -11,8 +10,8 @@ class ScheduledApp(BaseApp):
     group_by_field = 'app_connection_id'
 
     @property
-    def event_loader(self) -> ScheduledLoader:
-        return ScheduledLoader()
+    def event_loader(self):
```
jordanambra, line 13: What are the `event_loader` methods doing now?

corva/app/stream.py
```diff
@@ -59,7 +58,7 @@ def pre_process(self, context: StreamContext) -> None:
         context.event = event
 
     def post_process(self, context: StreamContext) -> None:
-        all_records: List[Record] = list(chain(*[subdata.records for subdata in context.event]))
+        all_records: List[StreamEventData.Record] = list(chain(*[subdata.records for subdata in context.event]))
```
jordanambra, line 61: Should this be in pre_process?

corva/middleware/loader.py
```diff
@@ -0,0 +1,14 @@
+from typing import Callable, Optional
+
+from corva.models.base import BaseContext
+
+
+def loader_factory(loader: Callable, loader_kwargs: Optional[dict] = None) -> Callable:
```
jordanambra, line 6: Can we simplify our interface to creating our own middleware functionality? I wouldn't want to have to implement the factory pattern every time. Can we make a simpler interface maybe?

corva/models/scheduled.py
```diff
@@ -35,3 +40,12 @@ class ScheduledEventData(BaseEventData):
     log_type: str
     log_identifier: Optional[str] = None
     day_shift_start: Optional[str] = None
+
+
+class ScheduledEvent(ListEvent[ScheduledEventData]):
+    @staticmethod
+    def from_raw_event(event: str, **kwargs) -> ScheduledEvent:
```
jordanambra, line 47: Overall, I like this strategy where the model is responsible better. Nice work.

corva/models/scheduled.py
```diff
@@ -35,3 +40,12 @@ class ScheduledEventData(BaseEventData):
...
+        parsed = parse_raw_as(List[List[ScheduledEventData]], event)  # type: List[List[ScheduledEventData]]
```
jordanambra, line 48: Do we need the type comment?

corva/models/scheduled.py
```diff
@@ -35,3 +40,12 @@ class ScheduledEventData(BaseEventData):
...
+        parsed = list(chain(*parsed))
```
jordanambra, line 49: Maybe leave a comment as to why we are chaining/flattening here?

corva/models/stream.py
```diff
@@ -1,27 +1,75 @@
...
+
+from corva.models.base import BaseContext, BaseEventData, ListEvent
 from corva.state.redis_state import RedisState
 
 
 class StreamContext(BaseContext):
     state: RedisState
 
 
+class RecordData(BaseEventData):
```
jordanambra, line 15: I don't think we can do this. `data` inside of a Record is variable, depending on the dataset.

corva/models/stream.py
```diff
@@ -1,27 +1,75 @@
...
+    def app_connection_id(self) -> int:
+        return self.metadata.apps[self.app_key].app_connection_id
+
+    @property
+    def app_stream_id(self) -> int:
+        return self.metadata.app_stream_id
+
+    @property
+    def is_completed(self) -> bool:
+        if len(self.records):
```
jordanambra, line 59:
```suggestion
        if self.records:
```

Any reason not to do this?

corva/models/stream.py
```diff
@@ -1,27 +1,75 @@
...
+            return self.records[-1].collection == 'wits.completed'
+
+        return False
+
+
+class StreamEvent(ListEvent[StreamEventData]):
+    @staticmethod
+    def from_raw_event(event: str, **kwargs) -> StreamEvent:
+        app_key: str = kwargs['app_key']
```
jordanambra, line 68: Can you also add app_version here? It should be an integer.

corva/models/task.py
```diff
@@ -44,3 +46,9 @@ class TaskEventData(BaseEventData):
     id: Optional[str] = None
     task_id: str
     version: conint(ge=2, le=2)  # only utils API v2 supported
+
+
+class TaskEvent(BaseEvent, TaskEventData):
```
jordanambra, line 51: Tasks can have arbitrary payloads that would be fitted into TaskEventData. Is that accounted for here?

jordanambra (changes requested):
Overall, this is looking a lot simpler in a good way. Once you get these changes in place I'd like to review again from scratch.

corva/middleware/loader.py
finomayato: can you please make PRs more completed? It gives a little insight into how functions are used. Reviewers cannot justify the implementation of something which is not used anywhere

corva/app/scheduled.py
oleksii-symon-corva-ai: ScheduledApp class will not be used anymore. Its logic will be extracted into middlewares and file deleted in [task](https://corvaqa.atlassian.net/browse/DC-754). I've changed the method, so tests don't fail.

corva/middleware/loader.py
oleksii-symon-corva-ai: yes, so the current interface looks like this
```python3
def middleware(context, call_next): pass
```

the factories are used only to parametrize the middleware, any other approach can be used e.g
```python3
# approach 1

class Middleware:
    def __init__(self, param1):
        self.param1 = param1

    def __call__(self, context, call_next):
        pass


mw = Middleware(param1='')  # parametrize like this


# approach 2

def middleware(context, call_next, param1):
    pass


mw = functools.partial(middleware, param1='')  # parametrize like this


# approach 3

def middleware_factory(param1):
    def middleware(context, call_next):
        pass

    return middleware


mw = middleware_factory(param1='')  # parametrize like this
```

corva/models/scheduled.py
oleksii-symon-corva-ai: thanks, deleted

corva/models/scheduled.py
oleksii-symon-corva-ai: done

corva/models/stream.py
oleksii-symon-corva-ai: I've accounted for missed, forgotten or added in future fields by setting `BaseEventData.Config.extra = allow` [link](https://github.com/corva-ai/python-sdk/blob/a715bf4d4cffe06f1794c20fc6e6ddc83c96a8ae/corva/models/base.py#L37)

corva/models/stream.py
oleksii-symon-corva-ai: Does it come from kafka or should be manually set and passed in kwargs?

corva/models/task.py
oleksii-symon-corva-ai: yes, `TaskEventData` inherits from `BaseEventData` which allows arbitrary fields [link](https://github.com/corva-ai/python-sdk/blob/a715bf4d4cffe06f1794c20fc6e6ddc83c96a8ae/corva/models/base.py#L37)

corva/app/stream.py
oleksii-symon-corva-ai: I think `last_processed_timestamp` and `last_processed_depth` should be extracted and saved in `post_process`. so in case stream task fails in `process`, wrong data is not saved to redis and failed task can be restarted with previous values

oleksii-symon-corva-ai: also this class won't be used anymore

corva/middleware/loader.py
finomayato, line 6: let's call this `get_loader` or `get_loader_fn`. It's basically what this function does. Also, it's not clear for me what loader does

oleksii-symon-corva-ai: it will be deleted in pr #14

finomayato: why do we review functionality which does not exist in the next PR? It does not make sense

oleksii-symon-corva-ai: branch in pr #14 was created from this one, so it had this functionality. it just turned out in process of development of pr #14 that we can get rid of this file

corva/models/scheduled.py
```diff
@@ -1,7 +1,12 @@
+from __future__ import annotations
+
 from datetime import datetime
-from typing import Optional
+from itertools import chain
+from typing import List, Optional
+
+from pydantic import Field, parse_raw_as
```
jordanambra, line 7: It's better practice to import classes and modules instead of functions directly.

`import itertools`
`import pydantic`

corva/models/stream.py
jordanambra: From kafka

jordanambra (approved):
I am approving this as-is, but DC-752, related to a new way of doing middleware, may need to come in front of this, and this may require refactoring, depending on our decisions there. Please don't merge this until that conversation is resolved.

corva/models/stream.py
oleksii-symon-corva-ai: added the field [here](https://github.com/corva-ai/python-sdk/blob/e5e2a5b34c3fb681c094f605600575edaf22d30f/corva/models/stream.py#L33), I assume it's optional as it is not documented anywhere

corva/models/scheduled.py
oleksii-symon-corva-ai: thanks, we would have to refactor whole project to fix all such issues

corva/models/stream.py
```diff
@@ -1,27 +1,76 @@
-from typing import Any, Dict, List, Optional
+from __future__ import annotations
 
-from corva.models.base import BaseContext, BaseEventData
+from typing import Dict, List, Optional
+
+from pydantic import parse_raw_as
```
jordanambra, line 5: Let's `import pydantic` and then call `pydantic.parse_raw_as`

jordanambra (approved):

PR #14, opened 2020-12-28
Feature/dc 753 add stream app type to Corva

oleksii-symon-corva-ai:
[DC-753](https://corvaqa.atlassian.net/browse/DC-753)

- added `stream` app type to `Corva`
- added new properties to `BaseContext`
- deleted default parameters from `Api` and `RedisAdapter`
- used pydantic.BaseSettings class to control env vars
- fixed `StreamEvent` and `ScheduledEvent` types to be objects instead of list of objects
- added tests for stream app


jordanambra (changes requested):
I'm about halfway through this review, and it seems like it has some crossover with another PR. Maybe we should get that PR finished and merged into here so we see only relevant changes?

corva/models/base.py
```diff
@@ -1,66 +1,135 @@
...
+    def api(self) -> Api:
+        if self._api is None:
+            kwargs = dict(
+                api_url=self.api_url,
+                data_api_url=self.api_data_url,
+                api_key=self.api_key,
+                api_name=self.api_app_name
+            )
+
+            if self.api_timeout is not None:
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: thanks, that was wrong attribute usage

corva/models/base.py
```diff
@@ -1,66 +1,135 @@
...
+    @property
+    def event(self) -> BaseEventTV:
+        if self._event is None:
+            self._event = BaseEventTV.from_raw_event(self.raw_event, app_key=self.app_key)
+
+        return self._event
+
+    @property
+    def api(self) -> Api:
+        if self._api is None:
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: fixed the issue with `functools.cached_property`

corva/models/base.py
```diff
@@ -1,66 +1,135 @@
...
+            if self.api_timeout is not None:
+                kwargs['max_retries'] = self.api_max_retries
+
+            self._api = Api(**kwargs)
+
+        return self._api
+
+    @property
+    def state(self) -> RedisState:
+        if self._state is None:
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: replaced `property` with `functools.cached_property`

corva/models/base.py
```diff
@@ -1,66 +1,135 @@
...
+            adapter_params = dict(
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: replaced with literal

corva/models/base.py
```diff
@@ -1,66 +1,135 @@
...
+
+        return self._state
 
-class BaseEventData(BaseModel):
-    class Config:
-        extra = Extra.allow
-        allow_population_by_field_name = True
+    @property
+    def state_data(self) -> BaseDataTV:
+        if self._state_data:
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: there is no if in update

oleksii-symon-corva-ai:
> I'm about halfway through this review, and it seems like it has some crossover with another PR. Maybe we should get that PR finished and merged into here so we see only relevant changes?

This branch doesnt have a crossover. All branches were created sequentially from one another and will be merged one after another. See at the top, this branch is compared to `feature/DC-756_migrate-loaders-logic-to-middlewares`. So it can be reviewed.

oleksii-symon-corva-ai:
> I'm about halfway through this review, and it seems like it has some crossover with another PR. Maybe we should get that PR finished and merged into here so we see only relevant changes?

@jordanambra, this branch doesnt have a crossover. All branches were created sequentially from one another and will be merged one after another. See at the top, this branch is compared to `feature/DC-756_migrate-loaders-logic-to-middlewares`. So it can be reviewed.

corva/models/base.py
```diff
@@ -1,47 +1,106 @@
...
+
+    @property
+    def cache_key(self) -> str:
+        return (
+            f'{self.settings.PROVIDER}/well/{self.event.asset_id}/stream/{self.event.app_stream_id}/'
+            f'{self.settings.APP_KEY}/{self.event.app_connection_id}'
+        )
+
+    @cached_property
+    def event(self) -> BaseEventTV:
```
oleksii-symon-corva-ai, line 67: needed for filtering the event in subclasses

corva/models/stream.py
```diff
@@ -62,14 +59,38 @@ def is_completed(self) -> bool:
...
+
+
+class StreamContext(BaseContext[StreamEvent, StreamStateData]):
+    cache_data_cls: Type[StreamStateData] = StreamStateData
+    filter_by_timestamp: bool = False
+    filter_by_depth: bool = False
+
+    @cached_property
+    def event(self) -> StreamEvent:
+        from corva.utils import FilterStreamEvent
```
finomayato, line 82: let's move this import to the top

corva/models/stream.py
```diff
@@ -62,14 +59,38 @@ def is_completed(self) -> bool:
...
+    last_processed_timestamp: int = -1
+    last_processed_depth: float = -1
+
+
+class StreamContext(BaseContext[StreamEvent, StreamStateData]):
+    cache_data_cls: Type[StreamStateData] = StreamStateData
+    filter_by_timestamp: bool = False
+    filter_by_depth: bool = False
+
+    @cached_property
```
finomayato, line 80: why it should be cached property? please, don't overuse caching, since it will be hard to reproduce and debug

tests/conftest.py
```diff
@@ -44,9 +46,93 @@ def redis(redis_adapter):
...
+    with open(DATA_PATH / 'stream_event.json') as stream_event:
+        return stream_event.read()
 
 
 class ComparableException(Exception):
     def __eq__(self, other):
         return type(self) is type(other) and self.args == other.args
+
+
+class StreamDataMixer:
```
finomayato, line 88: can we simplify the test's logic and not create classes with a specific logic of how tests should work?
It would be much simpler to understand what is happening in the test inside the scope of itself, rather than figure out the logic defined elsewhere

finomayato (changes requested):
* Overuse of caching
* Tests optimized for writing, not for reading (defined abstract classes with logic which should simplify the writing of tests, but complicate reading and changing)

corva/application.py
```diff
@@ -25,13 +30,71 @@ def __init__(self, middleware: Optional[List[Callable]] = None):
...
 
         return middleware_stack
 
     def add_middleware(self, func: Callable) -> None:
         self.user_middleware.append(func)
+
+    def stream(
+         self,
+         func=None,
+         *,
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: It would be better to pass parameters by name to not shuffle them. https://softwareengineering.stackexchange.com/a/384703

corva/models/base.py
```diff
@@ -1,47 +1,106 @@
...
+    cache_data_cls: Optional[Type[BaseDataTV]] = None
+
+    @property
+    def cache_key(self) -> str:
+        return (
+            f'{self.settings.PROVIDER}/well/{self.event.asset_id}/stream/{self.event.app_stream_id}/'
+            f'{self.settings.APP_KEY}/{self.event.app_connection_id}'
+        )
+
+    @cached_property
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: this allows to do filtering once [here](https://github.com/corva-ai/python-sdk/blob/687b6ebaa97fc8905f8e43db23009b65ba809a42/corva/models/stream.py#L100)

corva/models/base.py
```diff
@@ -1,47 +1,106 @@
...
+    def event(self) -> BaseEventTV:
+        return self._event
+
+    @cached_property
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: it is better to use one Api instance as it uses requests.Session which uses HTTP persistent connection [here](https://requests.readthedocs.io/en/master/user/advanced/)

corva/middleware/stream.py
```diff
@@ -0,0 +1,38 @@
...
+        default=StreamStateData.__fields__['last_processed_depth'].default
+    )
+
+    object.__setattr__(
+        context,
+        'cache_data',
+        StreamStateData(
+            last_processed_timestamp=last_processed_timestamp, last_processed_depth=last_processed_depth
+        )
+    )
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: deleted

corva/models/stream.py
oleksii-symon-corva-ai: deleted utils file, there were circular deps

corva/models/stream.py
oleksii-symon-corva-ai: cached_property was used to run filtering only once

tests/conftest.py
oleksii-symon-corva-ai: Stream events have many different fields. Without StreamDataMixer it'd be hard to create custom StreamEvent object in every test

corva/application.py
```diff
@@ -25,13 +30,71 @@ def __init__(self, middleware: Optional[List[Callable]] = None):
...
+         middleware: Optional[List[Callable]] = None,
+         tail_middleware: Optional[List[Callable]] = None
     ) -> List[Callable]:
         middleware = middleware or []
+        tail_middleware = tail_middleware or []
 
-        middleware_stack = middleware + self.user_middleware
+        middleware_stack = middleware + self.user_middleware + tail_middleware
 
         return middleware_stack
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: deleted the function as it really doesnt make much sense right now

corva/application.py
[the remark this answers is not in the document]
finomayato: In the same answer there are pitfalls:
> * It's yet another feature for new team members to learn when they're coming up to speed on Python and your codebase.
> * Extra complexity and distraction in the function definition. (What's that *?)
> * Some dev tools might not handle it well: syntax highlighters, diff tools, refactoring editors, code inspectors, C interop, RPC (remote procedure call) code generators, ...

We are implementing a user-facing library (an SDK). Because of that, I consider these pitfalls important enough to get rid of `*`. I'm biased toward engineering simplicity. You use your best judgment

corva/models/stream.py
finomayato: using local imports to resolve circular dependencies is a short-term fix. It will backfire later if we don't build the structure of the project right

tests/conftest.py
finomayato: Maybe it would be harder to create, but it would be easier to read and change. Engineers read code more than write. We need to worry about us reading this code, not how hard it would be to write it

corva/models/base.py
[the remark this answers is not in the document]
finomayato: This method does not ensure that one Api instance is used. You need to implement singleton on Api to achieve that. Not cache the result of the method. This is a wrong tool for singleton

corva/models/base.py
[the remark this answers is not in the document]
finomayato: Then it should be implemented just there. And this function will always give the same result. Is it something we want in the same runtime?

What happens without caching the result?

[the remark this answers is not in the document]
finomayato: If nothing happens - we don't need it. It's overoptimization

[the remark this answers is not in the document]
oleksii-symon-corva-ai: deleted the property

corva/models/base.py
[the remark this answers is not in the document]
oleksii-symon-corva-ai: `Api` class is not a singleton by itself. Really `@cached_property` servers here as a method of writing next code:

```python3
from pydantic import BaseModel, validator
from functools import cached_property
from typing import Optional


class Api:
    pass


# method 1
class Data(BaseModel):
    class Config:
        keep_untouched = (cached_property,)

    @cached_property
    def api(self) -> Api:
        return Api()


# method 2
class Data(BaseModel):
    class Config:
        arbitrary_types_allowed = True

    api_: Optional[Api] = None

    @property
    def api(self) -> Api:
        if self.api_ is not None:
            return self.api_

        self.api_ = Api()
        return self.api_


# method 3
class Data(BaseModel):
    class Config:
        arbitrary_types_allowed = True

    api: Api = None

    @validator('api', pre=True, always=True)
    def set_api(cls, v, values) -> Api:
        # extract Api kwargs from values here and pass to constructor

        return Api()
```

corva/models/stream.py
oleksii-symon-corva-ai: deleted event cached_property

tests/conftest.py
oleksii-symon-corva-ai: Q What current mixer logic does?
A It constructs and returns a pydantic model, providing required parameters to constuctor (without those params model raises an exception). We can override parameters as needed, otherwise just call the function without arguments to get the model.

Q Why do we need to generate models?
A We need to generate models (such as events) and cast them to json, because this is what will be received in real-world from kafka

Cool scenario would be the support of pydantic models in `mixer` [library](https://pypi.org/project/mixer/), so there is no need to write classes like this. But there is no support.

We are left with a few options then:

1. Delete the mixer from repository. Insert raw events in every test, modifying some fields as needed. Raw events look like this: `'[{"app_key": null, "records": [{"timestamp": null, "asset_id": 0, "company_id": 0, "version": 0, "measured_depth": null, "collection": "wits.completed", "data": {"hole_depth": null, "weight_on_bit": null, "state": null}}], "metadata": {"app_stream_id": 0, "source_type": null, "apps": {"provider.app-name": {"app_connection_id": 0}}}, "asset_id": 0}]'`.
Concern - it may be hard to maintain such raw events (in case model changes).

2. Use some library like [factory_boy](https://factoryboy.readthedocs.io/en/stable/), which would require creating factory_boy models (replicating pydantic models).

3. Current mixer class.

corva/application.py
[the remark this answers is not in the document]
oleksii-symon-corva-ai: I will leave it for now for following reasons.

1. it can always be deleted, without affecting backward compatibility
2. it helps not to shuffle many arguments provided by the decorator (write `@stream(filter_by_timestamp=True` and `@stream(filter_by_depth=True)` vs `@stream(True, False)`, `@stream(False, True)`)
2. it is probably time for devs to learn about this feature as it was added in python 3.0

corva/models/base.py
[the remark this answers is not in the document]
finomayato: I don't understand what are you trying to say? Are these methods - alternative solutions?

tests/conftest.py
finomayato: The example you provided is not big

corva/models/base.py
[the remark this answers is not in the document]
finomayato: why not make Api class a singleton?

corva/application.py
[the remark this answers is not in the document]
finomayato: I was wrong. It actually very useful here

corva/models/base.py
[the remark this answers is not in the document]
oleksii-symon-corva-ai: Yes, these methods are some alternatives.

I was trying to say, that using [functools.cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) saves you from burden of writing complex code with hidden private attributes, that support `@properties`. So, it simplifies code of our BaseContext class.

Simpler example would be:

```python3
from functools import cached_property


class Foo1:
    """cached_property doesnt require you to define and manage private attributes"""

    @cached_property
    def baz(self):
        return []


class Foo2:
    """does the same thing as Foo1"""

    def __init__(self):
        self._baz = []

    @property
    def baz(self):
        return self._baz

    @baz.setter
    def baz(self, value):
        self._baz = value


class Foo3:
    """Fails to achieve the same thing as Foo1"""

    @property  # cant do the same code as in Foo1
    def baz(self):
        return []


foo1 = Foo1()
foo2 = Foo2()
foo3 = Foo3()

foo1.baz.append(1)
foo2.baz.append(1)
foo3.baz.append(1)

assert foo1.baz == [1]
assert foo2.baz == [1]
assert foo3.baz == []


foo1.baz = []
foo2.baz = []

assert foo1.baz == []
assert foo2.baz == []
```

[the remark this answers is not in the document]
oleksii-symon-corva-ai: > why not make Api class a singleton?

Api class instances can be parametrized by passing different parameter sets to constructor. This parametrization requires different instances to be returned. Singleton would always return only one instance, which was the very first one created.

If we needed to optimize request.Session usage inside Api class, it would be more appropriate to use Flyweight pattern, which would return same instances for same sets of parameters.

These optimizations are not needed at this point in my opinion.

tests/conftest.py
oleksii-symon-corva-ai: what example?

if I understand correctly, you say, that option 1 would be the best?

oleksii-symon-corva-ai: will rewrite the tests as in option #1

corva/models/base.py
```diff
@@ -1,47 +1,87 @@
...
 class BaseEvent(ABC):
     @staticmethod
     @abstractmethod
-    def from_raw_event(event: str, **kwargs) -> BaseEvent:
+    def from_raw_event(event: str, **kwargs) -> Union[List[BaseEvent], BaseEvent]:
+        pass
+
+
+class BaseData(BaseModel):
+    class Config(BaseConfig):
```
finomayato, line 30: is there any chance we can void using nested classes?

corva/models/base.py
```diff
@@ -1,47 +1,106 @@
...
-    class Config:
-        extra = Extra.allow
-        allow_population_by_field_name = True
 
+class BaseContext(GenericModel, Generic[BaseEventTV, BaseDataTV]):
+    """Stores common data for running a Corva app."""
 
-BaseEventDataTV = TypeVar('BaseEventDataTV', bound=BaseEventData)
+    class Config(BaseConfig):
+        pass
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: it is needed to inherit common setting from base class

corva/models/stream.py
```diff
@@ -63,14 +78,52 @@ def is_completed(self) -> bool:
...
     @staticmethod
-    def from_raw_event(event: str, **kwargs) -> StreamEvent:
-        app_key: str = kwargs['app_key']
+    def filter(
+         event: StreamEvent, by_timestamp: bool, by_depth: bool, last_timestamp: int, last_depth: float
+    ) -> StreamEvent:
+        records = event.records
+
+        if event.is_completed:
+            records = records[:-1]  # remove "completed" record
```
finomayato, line 100: we can have only 1 completed record and it will be always at the end of the list?

corva/models/stream.py
```diff
@@ -63,14 +78,52 @@ def is_completed(self) -> bool:
...
+                continue
+
+            new_records.append(record)
+
+        return event.copy(update={'records': new_records}, deep=True)
+
+
+class StreamStateData(BaseData):
+    last_processed_timestamp: int = -1
+    last_processed_depth: float = -1
```
finomayato, line 116: why not `None`?

corva/models/base.py
oleksii-symon-corva-ai: we can do
```suggestion
    class BaseData(BaseModel):
        Config = BaseConfig
```

corva/models/stream.py
oleksii-symon-corva-ai: thanks, changed to `None`

corva/models/stream.py
ville: Yes

corva/models/base.py
oleksii-symon-corva-ai: resolved the issue using base classes [CorvaBaseModel](https://github.com/corva-ai/python-sdk/blob/911358315edce6151fbbb48faea021a7460488db/corva/models/base.py#L29)

corva/models/stream.py
finomayato: probably need to document that here, because it is not clear when reading this code

finomayato (approved):
Go ahead. We already did a lot of work on it and we need to move forward

corva/settings.py
```diff
@@ -1,15 +1,40 @@
...
+
+    @property
+    def PROVIDER(self) -> str:
+        if provider := getenv('PROVIDER') is not None:
+            return provider
+
+        return self.APP_KEY.split('.')[0]
+
+
+CORVA_SETTINGS = CorvaSettings()
```
jordanambra, line 40: Can we call this SETTINGS instead of CORVA_SETTINGS?

corva/models/task.py
```diff
@@ -52,3 +47,7 @@ class TaskEvent(BaseEvent, TaskEventData):
     @staticmethod
     def from_raw_event(event: str, **kwargs) -> TaskEvent:
         return pydantic.parse_raw_as(TaskEvent, event)
+
+
+class TaskContext(BaseContext[TaskEvent, CorvaBaseModel]):
```
jordanambra, line 52: Why is this inherited like this?

corva/models/stream.py
```diff
@@ -57,20 +72,61 @@ def app_stream_id(self) -> int:
...
+    def filter(
+         event: StreamEvent, by_timestamp: bool, by_depth: bool, last_timestamp: int, last_depth: float
+    ) -> StreamEvent:
+        records = event.records
+
+        if event.is_completed:
+            # there can only be 1 completed record, always located at the end of the list
+            records = records[:-1]  # remove "completed" record
+
+        new_records = []
```
jordanambra, line 105: I think this should be a generator

corva/application.py
```diff
@@ -1,2 +1,64 @@
+from typing import Any, Callable, List, Optional
+
+from corva.models.stream import StreamContext, StreamEvent
+from corva.network.api import Api
+from corva.settings import CorvaSettings, CORVA_SETTINGS
```
jordanambra, line 5: This is confusing. I'd like to rename CORVA_SETTINGS to SETTINGS

corva/application.py
```diff
@@ -1,2 +1,64 @@
...
 class Corva:
-    pass
+    def __init__(
+         self,
+         api_url: Optional[str] = None,
+         data_api_url: Optional[str] = None,
+         cache_url: Optional[str] = None,
+         api_key: Optional[str] = None,
+         app_key: Optional[str] = None,
+         api_timeout: Optional[int] = None,
```
jordanambra, line 17: I think since this is an external interface, we should have better names here. Here are my preferences:

* remove data_api_url: I'm also not sure that we need to be able to override the data API URL here.
* api_timeout => timeout
* api_max_retries => max_retries
* cache_kwargs => cache_settings

corva/application.py
```diff
@@ -1,2 +1,64 @@
...
+         event: str,
+         *,
+         filter_by_timestamp: bool = False,
+         filter_by_depth: bool = False
+    ) -> List[Any]:
+        events = StreamEvent.from_raw_event(event=event, app_key=self.settings.APP_KEY)
+
+        results = []
+
+        for event in events:
```
jordanambra, line 52: Can this be a generator?

corva/models/base.py
```diff
@@ -1,47 +1,86 @@
...
+    @property
+    def cache_key(self) -> str:
+        return (
+            f'{self.settings.PROVIDER}/well/{self.event.asset_id}/stream/{self.event.app_stream_id}/'
+            f'{self.settings.APP_KEY}/{self.event.app_connection_id}'
+        )
+
+    @property
+    def cache(self) -> RedisState:
+        if self._cache is not None:
```
jordanambra, line 64: Why not `if self._cache`?

corva/models/base.py
```diff
@@ -1,47 +1,86 @@
...
+
+    event: BaseEventTV
+    settings: CorvaSettings
+    api: Api
+    _cache: Optional[RedisState] = None
+
     user_result: Any = None
 
+    # cache params
+    cache_kwargs: dict = {}
```
jordanambra, line 52: Can we rename this to cache_settings?

corva/models/base.py
```diff
@@ -1,47 +1,86 @@
...
 
+        self._cache = RedisState(redis=redis_adapter)
 
-BaseEventDataTV = TypeVar('BaseEventDataTV', bound=BaseEventData)
+        return self._cache
 
+    @property
+    def cache_data(self) -> CorvaBaseModelTV:
+        state_data_dict = self.cache.load_all()
+        return self.cache_data_cls(**state_data_dict)
```
jordanambra, line 80: I think we need to check if cache_data_cls exists, or otherwise make the class non-optional.

corva/models/scheduled.py
```diff
@@ -42,13 +37,17 @@ class ScheduledEventData(BaseEventData):
     day_shift_start: Optional[str] = None
 
 
-class ScheduledEvent(ListEvent[ScheduledEventData]):
+class ScheduledEvent(BaseEvent, ScheduledEventData):
```
jordanambra, line 40: Why do we have scheduled event updates in this PR?

corva/models/stream.py
```diff
@@ -28,24 +23,44 @@ class Record(BaseEventData):
...
+         after filtering, we may end up with empty records. Which leads to asset_id becoming None.
+         Using this validator we are able to dynamically set and store value of asset_id,
+         no matter what happens to records.
+        """
+
+        records = values['records']  # type: List[Record]
+        if records:
+            return records[0].asset_id
+
+        raise ValueError('Can\'t determine asset_id as records are empty (which should not happen).')
```
jordanambra, line 63: This should be double quotes, since you have a single quote in the sentence.

corva/models/stream.py
```diff
@@ -28,24 +23,44 @@ class Record(BaseEventData):
...
+    source_type: Optional[str] = None
     apps: Dict[str, AppMetadata]
 
 
-class StreamEventData(BaseEventData):
+class StreamEventData(CorvaBaseModel):
     app_key: Optional[str] = None
     records: List[Record]
     metadata: StreamEventMetadata
+    asset_id: int = None
```
jordanambra, line 41: Optional[int]?

corva/models/stream.py
```diff
@@ -28,24 +23,44 @@ class Record(BaseEventData):
     data: RecordData
 
 
-class AppMetadata(BaseEventData):
+class AppMetadata(CorvaBaseModel):
     app_connection_id: int
     app_version: Optional[int] = None
```
jordanambra, line 28: I think we can make this not Optional. We want this all the time, and anybody who is using the SDK will also be using Dev Center or otherwise deploying with Packager, which should help for this.

corva/models/stream.py
```diff
@@ -28,24 +23,44 @@ class Record(BaseEventData):
...
+        def asset_id(self) -> Optional[int]:
+            return self.records[0].asset_id if self.records else None
+
+        The issue with the above method is:
+         after filtering, we may end up with empty records. Which leads to asset_id becoming None.
+         Using this validator we are able to dynamically set and store value of asset_id,
+         no matter what happens to records.
+        """
+
+        records = values['records']  # type: List[Record]
```
jordanambra, line 59: This could probably be a try/except instead, for speed.

corva/network/api.py
```diff
@@ -6,23 +6,23 @@
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
-from corva import settings
-
 
 class Api:
     ALLOWED_METHODS = {'GET', 'POST', 'PATCH', 'PUT', 'DELETE'}
+    DEFAULT_TIMEOUT = 600
```
jordanambra, line 12: Shouldn't these go into settings?

corva/settings.py
```diff
@@ -1,15 +1,40 @@
...
-API_KEY: Final[str] = getenv('API_KEY')
+from pydantic import BaseSettings
 
-# Logger
-LOG_LEVEL: Final[str] = getenv('LOG_LEVEL', 'WARN')
-LOG_ASSET_ID: Final[int] = int(getenv('LOG_ASSET_ID', -1))
 
-# Storage
-CACHE_URL: Final[str] = getenv('CACHE_URL')
+class CorvaSettings(BaseSettings):
```
jordanambra, line 7: Let's change the name of this to Settings.

corva/stream.py
```diff
@@ -0,0 +1,46 @@
+from typing import Any, Callable
+
+from corva.models.stream import StreamContext, StreamEvent, StreamStateData
+
+
+def stream_runner(fn: Callable, context: StreamContext) -> Any:
+    if (last_timestamp := context.cache_data.last_processed_timestamp) is None:
```
jordanambra, line 7: Nice to see our friend Mr. Walrus :D

jordanambra (changes requested):

corva/settings.py
oleksii-symon-corva-ai: Named it `CORVA_SETTINGS` to avoid vague imports like: `from corva.settings import SETTINGS`, as pointed out in some previous PR

corva/settings.py
oleksii-symon-corva-ai: Named it `CorvaSettings` to avoid vague imports like: `from corva.settings import Settings`, as pointed out in some previous PR. If you think it's alright, will return it

corva/models/task.py
oleksii-symon-corva-ai: Good question, I've deleted second generic argument `CorvaBaseModel` from `BaseContext` at all, as it is needed only in StreamContext.

As for leftover `BaseContext[TaskEvent]` part, it is inherited like this because:

1. BaseContext is a generic class (it was made generic to get correct type hints in subclasses - `TaskContext` in this case)

corva/models/stream.py
oleksii-symon-corva-ai: Generators are better to use in simple cases. In this case it'd be complex smth like
```python
        new_records = [
            record
            for record in records
            if (
                 (by_timestamp and record.timestamp > last_timestamp)
                 or
                 (by_depth and record.measured_depth > last_depth)
                 or (not by_timestamp and not by_depth)
            )
        ]
```

corva/models/stream.py
oleksii-symon-corva-ai: Asked Ville about this field, here is what he said

> I think it should be optional since old apps might not have it

corva/settings.py
jordanambra: If that's an issue, you can change the module name to `configuration.py`. I think that resolves everything?

corva/models/stream.py
jordanambra: Old apps won't be using the Python SDK until they are updated though.

corva/models/stream.py
jordanambra: I only mean something like this, which could be called from `filter()`.

```suggestion
        def process_records(self, records):
            for record in records:
                if by_timestamp and record <= last_timestamp:
                    continue
                
                if by_depth and record.measured_depth <= last_depth:
                    continue
                    
                yield record
```

oleksii-symon-corva-ai: cool, I guess this will also work faster

corva/models/stream.py
oleksii-symon-corva-ai: made the field required

corva/models/stream.py
oleksii-symon-corva-ai: reworked the function, take a look

corva/network/api.py
oleksii-symon-corva-ai: Settings contain only env variables as described [here](https://www.notion.so/corva/Data-Apps-454ddb3fdb174e10b5cd53679ad999ca).
`DEFAULT_TIMEOUT` belongs to `Api` and is not meant to be changed.
User can set custom default timeout on `Api` initialization like `Api(timeout=601)`

corva/settings.py
oleksii-symon-corva-ai: Yes, renamed the file, class and global variable

corva/settings.py
oleksii-symon-corva-ai: Yes, renamed the file, class and global variable

corva/application.py
oleksii-symon-corva-ai: Yes, renamed the file, class and global variable

corva/configuration.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # misc
+    APP_KEY: Optional[str] = None  # <provider>.<app-name-with-dashes>
+
+    @property
+    def APP_NAME(self) -> str:
+        if app_name := getenv('APP_NAME') is not None:
+            return app_name
+
+        app_name_with_dashes = self.APP_KEY.split('.')[1]
```
jordanambra, line 27: This could cause an index error if the app key is mangled. Also, this should probably limit the splitting to 1 result.

jordanambra (changes requested):

corva/application.py
oleksii-symon-corva-ai: agreed
might as well delete, cache_url and api_url along with data_api_url

oleksii-symon-corva-ai: agreed
might as well delete all params that can be retreived from SETTINGS variable along with data_api_url
so we will be left with timeout, max_retries and cache_settings

corva/application.py
oleksii-symon-corva-ai: Yes, it can. But what is the point, if we will return a list (that holds all elements in memory) ?

corva/models/base.py
oleksii-symon-corva-ai: the function was moved to corva/models/stream.py, where the issue doesnt exist

corva/models/scheduled.py
oleksii-symon-corva-ai: I've deleted ListEvent in this PR, which would cause tests to fail, without updating ScheduledEvent

corva/models/stream.py
oleksii-symon-corva-ai: the single quote in the sentence is escaped, so this works fine

corva/models/stream.py
oleksii-symon-corva-ai: It is required to be int. the reason for such type hint: https://pydantic-docs.helpmanual.io/usage/validators/#validate-always
added comment to code

corva/models/base.py
oleksii-symon-corva-ai: fixed

corva/models/base.py
oleksii-symon-corva-ai: renamed

corva/configuration.py
oleksii-symon-corva-ai: true, but this is our responsibility to set correct env vars

PR #15, opened 2021-01-13
Feature/dc 754 add scheduled app type to Corva

oleksii-symon-corva-ai:
[DC-754](https://corvaqa.atlassian.net/browse/DC-754)

- added scheduled app
- added tests
- deleted useless files


finomayato:
name of this PR does not correspond to changes

corva/scheduled.py
```diff
@@ -0,0 +1,11 @@
+from typing import Any, Callable
```
finomayato, line 1: name of this module is not descriptive. What "scheduled"? `runners/scheduled.py` will be more descriptive, since without visiting the code of the file itself, you can tell that it is implementation of scheduled runner

finomayato (changes requested):
Look how great! Now to implement a new runner we need ~30 lines of code. Good job!

I'm requesting changes, since PR name does not correspond to the changes you've made

corva/scheduled.py
oleksii-symon-corva-ai: thanks, moved both files to corva/runners/

finomayato (approved):
Perfect

PR #16, opened 2021-01-15
Feature/minor logger and redis improvements

oleksii-symon-corva-ai:
- improved cache type hinting
- improved api type hinting
- deleted logger


corva/network/api.py
```diff
@@ -54,20 +54,25 @@ def _init_session(api_key: str, app_name: str, max_retries: int, allowed_methods
 
         return session
 
-    def get(self, path: str, **kwargs):
-        return self._request('GET', path, **kwargs)
+    @property
+    def get(self):
+        return self._request('GET')
```
finomayato, line 59: why are all methods are properties?

corva/network/api.py
```diff
@@ -78,35 +83,36 @@ def _get_url(self, path: str):
...
+        def _request_helper(
+             path: str,
+             *,
+             data: Optional[dict] = None,  # request body
+             params: Optional[dict] = None,  # url query string params
+             headers: Optional[dict] = None,  # additional headers to include in request
+             max_retries: Optional[int] = None,  # custom value for max number of retries
+             timeout: Optional[int] = None,  # request timeout in seconds
+        ) -> Response:
+            if method not in self.ALLOWED_METHODS:
```
finomayato, line 96: Overcomplicated. We can make do without a nested function

corva/state/redis_adapter.py
```diff
@@ -14,64 +11,52 @@ class RedisAdapter(Redis):
...
+        n_set = super().hset(name=self.name, key=key, value=value, mapping=mapping)
 
-        if expiry is None and self.pttl(name=name) > 0:
-            self.persist(name=name)
+        if expiry is None and self.pttl() > 0:
+            self.persist(name=self.name)
 
         if expiry is not None:
-            self.expire(name=name, time=expiry)
+            self.expire(name=self.name, time=expiry)
```
finomayato, line 39: need documentation on why do we re-define `hset` default behavior

corva/state/redis_adapter.py
```diff
@@ -14,64 +11,52 @@ class RedisAdapter(Redis):
...
-        name = name or self.default_name
-        return super().hdel(name, *keys)
+    def hdel(self, keys: List[str]) -> int:
+        return super().hdel(self.name, *keys)
 
-    def delete(self, *names: List[str]) -> int:
-        names = names or [self.default_name]
-        return super().delete(*names)
+    def delete(self) -> int:
+        return super().delete(self.name)
```
finomayato, line 53: Is there any way we can use a non-modified Redis adapter?

finomayato (changes requested):
Before making improvements I'd like to try out existing implementation. I think we are ready to work with bar `requests` and bare `Redis` adapter to build an application

corva/app/base.py
```diff
@@ -52,15 +47,13 @@ def _run(self, event: Event) -> None:
         try:
             context = self.get_context(event=event)
         except Exception:
```
jordanambra, line 49: These try/except blocks seem pointless now that we are just reraising.

corva/network/api.py
```diff
@@ -54,20 +54,25 @@ def _init_session(api_key: str, app_name: str, max_retries: int, allowed_methods
 
         return session
 
-    def get(self, path: str, **kwargs):
-        return self._request('GET', path, **kwargs)
+    @property
```
jordanambra, line 57: You should leave these as functions and call or pass parameters through directly instead of using a layer of indirection with properties and wrapped functions. See the requests library itself for an example, or just revert back to how it was.

jordanambra (changes requested):
I have some general questions here, as well as inline feedback on the PR:

1. What task does this correspond to?
2. Why are we refactoring the API interface?
3. Why are we removing the logger?

There are a lot of different changes all wrapped up into a single PR here. It would be nice to limit this to only API, only Redis, or only logger changes, if those are necessary. Can we justify this PR?

corva/state/redis_adapter.py
oleksii-symon-corva-ai: yes, we can. logic from overridden functions in RedisAdapter can be moved to corresponding functions in RedisState (e.g move logic from RedisAdapter.hset to RedisState.store). we will get rid of RedisAdapter class and logic will become more straightforward.

corva/network/api.py
oleksii-symon-corva-ai: They were made properties to get proper parameter information:

Before:
image


After:
image

oleksii-symon-corva-ai:
@finomayato 

> I think we are ready to work with bare `requests`

We can get rid of Api class and instead pass user a Session instance with binded Authorization headers. In this case users will have to write all timeouts and retries themselves. I think it will simplify the code and make the library more intuitive, what do you think?

corva/network/api.py
oleksii-symon-corva-ai: I didn't find another way to get proper parameter information like displayed above

corva/network/api.py
oleksii-symon-corva-ai: The properties were added with the purpose of getting proper parameter information in IDE

https://github.com/corva-ai/python-sdk/pull/16#discussion_r559421792

corva/network/api.py
finomayato: it does not sound like a good idea to use language features to force IDE to behave in the way you want it to.
Since `get` is a method of the `Api` class - it's ok to have it as the first parameter in parameters

corva/network/api.py
finomayato: make user-facing function accept these as parameters?

finomayato:
> We can get rid of Api class and instead pass user a Session instance with binded Authorization headers. In this case users will have to write all timeouts and retries themselves. I think it will simplify the code and make the library more intuitive, what do you think?

Possibly. But not for certain. We will know for sure when we implement the first app with this library. For now, we can just expose `AUTH_HEADERS`. Or create a really lightweight class `Api` with methods: `get` and `post` which basically has the next implementation:
```python
class Api:
   def get(self, url, **kwargs):
        headers = kwargs.pop('headers', None)
        if headers:
             headers['Token'] = f'Bearer {AUTH_TOKEN}'
        return requests.get(url, headers=headers, **kwargs)
```

The one and only important question is: how a user will interact with this library?

jordanambra:
Can we leave this PR on hold for now? I don't want to make significant interface changes on this, and I'd prefer that we focus just on the stream/scheduled app implementation until that's released.

oleksii-symon-corva-ai:
Yes, lets hold

yataghan:
This PR is no longer relevant. Closing it.

PR #17, opened 2021-01-19
Feature/dc 676 add system and readme docs

oleksii-symon-corva-ai:
[DC-676 [Python Worker] Add system and readme docs](https://corvaqa.atlassian.net/browse/DC-676)


README.md
```diff
@@ -1,24 +1,272 @@
...
+- [Installation](#installation)
+- [App types](#app-types)
+- [Event](#event)
+- [Api](#api)
+- [Cache](#cache)
+- [Contributing](#contributing)
+
+## Requirements
+
 Python 3.8+
```
finomayato, line 15: is there any chance that it will work on 3.7 too?

README.md
```diff
@@ -1,24 +1,272 @@
...
 
-# Local development
+- [Pydantic](https://github.com/samuelcolvin/pydantic)
+- [Redis](https://pypi.org/project/redis/)
+- [Requests](https://pypi.org/project/requests/)
+- [urllib3](https://pypi.org/project/urllib3/)
```
finomayato, line 20: Do users really need to know that?

README.md
```diff
@@ -1,24 +1,272 @@
...
+
+```console
+$ pip install corva-sdk
+```
+
+## App types
+
+There are three app types, that you can build: `stream`, `scheduled` and `task`.<br>
+**Note**: it is recommended to use type hints like in examples below,
+so that editors and tools can give you better support.
```
finomayato, line 32: This recommendation does not change the way how our library works. Let's believe that users choose using or not using type hints by themselves

README.md
```diff
@@ -1,24 +1,272 @@
...
+```
+
+`Corva.stream` provides two additional parameters:
+
+- `filter_by_timestamp` - if set to `True` will take the latest processed
+  `timestamp` from cache and filter out records from event with 
+  either smaller or same `timestamp`;
+- `filter_by_depth` - if set to `True` will take the latest processed
+  `measured_depth` from cache and filter out records from event with
+  either smaller or same `measured_depth`.
```
finomayato, line 60: as a user who faces this library for the first time - it's not clear to me what is it "Event", `measured_depth`

README.md
```diff
@@ -1,24 +1,272 @@
...
+def lambda_handler(event, context):
+    """AWS lambda handler"""
+    corva = Corva()  # 3 initialize Corva
+    corva.stream(stream_app, event)  # 4 run stream_app by passing it and event to Corva.stream
+```
+
+`Corva.stream` provides two additional parameters:
+
+- `filter_by_timestamp` - if set to `True` will take the latest processed
+  `timestamp` from cache and filter out records from event with
```
finomayato, line 56: do users need to know about cache at all?

README.md
```diff
@@ -1,24 +1,272 @@
...
+#### Stream
+
+```python
+from corva import Api, Cache, Corva, StreamEvent
+
+
+# 1 define a function with required parameters, that will be provided by sdk
+def stream_app(event: StreamEvent, api: Api, cache: Cache):
+    """Main logic function"""
+    pass
```
finomayato, line 43: let's do something here. "Hello World" code at least

README.md
```diff
@@ -1,24 +1,272 @@
...
+#### Task (will be added soon)
+
+```python
+from corva import Api, Corva, TaskEvent
+
+
+# note, that task app doesn't receive cache parameter
+def task_app(event: TaskEvent, api: Api):
+    pass
+
```
finomayato, line 86: Let's not have something in the documentation which does not exist. Is there any reason to keep it here?

README.md
```diff
@@ -1,24 +1,272 @@
...
+so that editors and tools can give you better support.
+
+#### Example:
+
+```python
+from corva import Api, Cache, Corva, StreamEvent
+
+
+def stream_app(event: StreamEvent, api: Api, cache: Cache):
+    event.records[0].measured_depth  # get measured depth of first event record
```
finomayato, line 109: As a developer, I will be interested in the data structure of the event, or class API

README.md
```diff
@@ -1,24 +1,272 @@
...
+
+
+def lambda_handler(event, context):
+    corva = Corva()
+    corva.stream(stream_app, event)
+```
+
+## Api
+
+The apps might need to communicate with Corva API and Corva Data API.
```
finomayato, line 119: does the user need to know the difference? If so - we need to describe these services

finomayato (commented):
* Notes about applying typehints are redundant
* Let's make examples to actually show something. So the user can copy-paste them and see that it works
* No section about local development
* User needs references on entity description when we mention them. For example, when describing additional parameters for `Corva.stream` we mention cache and `measured_depth`. At the time of reading it - I do not know what a cache is and what measured_depth mean

README.md
bpamos: Agreed. That would be nice.

bpamos: maybe a docstring following this format would explain things to an outsider.

`         """
        Prints the person's name and age.

        If the argument 'additional' is passed, then it is appended after the main info.

        Parameters
        ----------
        additional : str, optional
            More info to be displayed (default is None)

        Returns
        -------
        None
        """`

README.md
finomayato: 3.7 is still widely used. We don't know which version did user used when written his algorithms. How hard it will be to check if everything works for python 3.7?

jordanambra: We can potentially support Python 3.6, 3.7, and 3.8, as those are the available AWS Lambda runtimes. We probably should do at least 3.7 though, since 3.6 will be EOL at the end of 2021.

oleksii-symon-corva-ai: Created a task to add 3.7 support [link](https://corvaqa.atlassian.net/browse/DC-869)

README.md
oleksii-symon-corva-ai: added link to event section. we need to add a section describing event fields, but this should be done after deleting obsolete event fields. as for now lets leave it as it is

README.md
oleksii-symon-corva-ai: maybe we will generate docs from Api class automatically in the future. as for now it is simpler to not describe each function and parameters here as they might change soon and make docs useless

README.md
oleksii-symon-corva-ai: added links to docs for more info

README.md
oleksii-symon-corva-ai: i might need to develop a small app first like [this task](https://corvaqa.atlassian.net/browse/DC-675) to add a real hello world example

oleksii-symon-corva-ai:
> No section about local development

There is a section #Contributing

README.md
oleksii-symon-corva-ai: added an example - simplified version of what gamma-depth app does

README.md
finomayato: Ok

finomayato (approved):

PR #19, opened 2021-02-01
Feature/dc 906 pull api key from client context

oleksii-symon-corva-ai:
[DC-906 Pull API key from ClientContext](https://corvaqa.atlassian.net/browse/DC-906)

- now api_key is looked for in lambda context first. if it is not present in the context - env var is scanned. if not found in both palces - an error is raised 
- fixed small bugs detected by mypy


corva/application.py
```diff
@@ -11,23 +12,31 @@
...
         params:
+         context: AWS Lambda context object
          timeout: api request timeout, set None to use default value
          max_retries: number or api retries for failed request, set to None to use default value
          cache_settings: additional cache settings
         """
 
+        client_context = CorvaLambdaClientContext.from_context(context)
+
+        if (api_key := client_context.api_key or SETTINGS.API_KEY) is None:
```
finomayato, line 31: maybe let's avoid new syntax, since we want this library to work with older python versions?

corva/models/context.py
```diff
@@ -0,0 +1,13 @@
...
+
+import pydantic
+
+
+class CorvaLambdaClientContext(pydantic.BaseModel):
+    @classmethod
+    def from_context(cls, context: Any) -> CorvaLambdaClientContext:
+        return cls(api_key=getattr(context.client_context, 'api_key', None))
+
+    api_key: Optional[str] = None
```
finomayato, line 13: what is the purpose of this class?

finomayato (changes requested):
`CorvaLambdaClientContext` model seems redundant. Conversation required

corva/models/context.py
oleksii-symon-corva-ai: the only reason for it to exist, is to validate the api_key type

corva/application.py
oleksii-symon-corva-ai: fixed

corva/models/context.py
finomayato: what kind of validation does this model do? I see that the `api_key` is optional and should be a string. Is this all the validation?

oleksii-symon-corva-ai: yes it is, we can probably delete the model

finomayato (approved):

ville (approved):

PR #20, opened 2021-02-03
Api class improvements

oleksii-symon-corva-ai:
[DC-916 Rework Api class to inherit from requests.Session](https://corvaqa.atlassian.net/browse/DC-916)

After conversation with Vadim, we agreed that it'd would be better to use composition instead of inheritance. So this pull request mostly includes Api improvements.

- reworked Api class
  - added usage of bare requests.request instead of requests.Session
  -  deleted retries
  - set lower default timeout
  - fixed [DC-757 Fix Api._get_url method on Windows](https://corvaqa.atlassian.net/browse/DC-757) along the way
  - changed regex patters for data api
  - completed [DC-698 Rework Api tests](https://corvaqa.atlassian.net/browse/DC-698) along the way
- moved Api class to corva/api.py from corva/network/api.py
- not related to task: added test to test_api_key.py


oleksii-symon-corva-ai:
Will fix the tests if the idea is validated

corva/network/api.py
```diff
@@ -1,134 +1,111 @@
...
-            'Authorization': f'API {api_key}',
-            'X-Corva-App': app_name
-        })
-        session.mount(
+        self.mount(
             'https://',
-            HTTPAdapter(
-                max_retries=Retry(
-                    total=max_retries,
+            requests.adapters.HTTPAdapter(
```
oleksii-symon-corva-ai, line 47: maybe leave it to users to implement retries?

finomayato: yep

finomayato (approved):

corva/api.py
```diff
@@ -0,0 +1,114 @@
+import posixpath
+import re
+from typing import Optional
+
+import requests
+
+
+class Api:
+    """Provides a convenient way to access Corva API and Corva Data API
```
jordanambra, line 9: Please fix indentation and spacing in the docstring here

corva/api.py
```diff
@@ -0,0 +1,114 @@
...
+        timeout: Optional[int] = None,
+    ):
+        self.api_url = api_url
+        self.data_api_url = data_api_url
+        self.api_key = api_key
+        self.app_name = app_name
+        self.timeout = timeout or self.TIMEOUT
+
+    @property
+    def auth_headers(self):
```
jordanambra, line 33: This should be renamed to `headers` or `standard_headers`.

corva/api.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if suffix.startswith('http'):
+            return suffix
+
+        suffix = suffix.lstrip(
+            '/'
+        )  # delete leading forward slash for posixpath.join to work correctly
+
+        # search text like api/v1 or api/v10 in path
+        if bool(re.search(r'api/v\d+', suffix)):
```
jordanambra, line 71: Do we have some unit tests to test the cases here? I can provide a list of URLs and which API they should point to, but I am not sure this properly matches.

corva/api.py
```diff
@@ -0,0 +1,114 @@
...
+    def patch(self, path: str, **kwargs):
+        return self._request('PATCH', path, **kwargs)
+
+    def put(self, path: str, **kwargs):
+        return self._request('PUT', path, **kwargs)
+
+    def delete(self, path: str, **kwargs):
+        return self._request('DELETE', path, **kwargs)
+
+    def _get_url(self, suffix: str):
```
jordanambra, line 54: `suffix` should probably be renamed `path`. `suffix` has a bit of a different meaning here that isn't clearly understood at first.

corva/api.py
```diff
@@ -0,0 +1,114 @@
...
+        timeout = timeout or self.timeout
+
+        headers = {
+            **self.auth_headers,
+            **(headers or {}),
+        }
+
+        response = requests.request(
+            method=method,
+            url=self._get_url(path),
```
jordanambra, line 107: I'd like to pull the return value of this into a variable in the _request method so that it's part of locals() and is reported to Rollbar if we have issues. Basically just do the same thing you did with `headers`.

corva/api.py
```diff
@@ -0,0 +1,114 @@
...
+         method: HTTP method
+         path: url to call
+         data: request body, that will be casted to json
+         params: url query string params
+         headers: additional headers to include in request
+         timeout: custom request timeout in seconds
+        returns: requests.Response instance
+        """
+
+        timeout = timeout or self.timeout
```
jordanambra, line 98: Let's put bounds on this `timeout` value. It should be between 3 and 30 seconds.

tests/test_api.py
```diff
@@ -0,0 +1,100 @@
...
+    api = Corva(context=None).stream(app, event)[0]
+
+    api.get('')  # do some api call
+
+    request_patch.assert_called_once()
+    assert 'Authorization' in request_patch.call_args.kwargs['headers']
+    assert 'X-Corva-App' in request_patch.call_args.kwargs['headers']
+
+
+@pytest.mark.parametrize(
```
jordanambra, line 34: This is a typo here right?

jordanambra (changes requested):

corva/api.py
jordanambra: Can we add the current app's version here, and pass it in as `X-Corva-App-Version`?

oleksii-symon-corva-ai: yes, where do I take it from

corva/api.py
oleksii-symon-corva-ai: we have tests, take a look https://github.com/corva-ai/python-sdk/blob/8893009db43743df311eccd3416279338b6b8e29/tests/test_api.py#L44

corva/api.py
oleksii-symon-corva-ai: renamed

tests/test_api.py
oleksii-symon-corva-ai: ci (tests) would fail if there was a typo

jordanambra: LOL how did I never notice this misspelling in pytest? Ridiculous.

corva/api.py
jordanambra: Good question. Let me make a new task to deal with that, it's a bigger issue.

corva/api.py
```diff
@@ -0,0 +1,129 @@
...
+            **(headers or {}),
+        }
+
+        response = requests.request(
+            method=method,
+            url=url,
+            params=params,
+            json=data,
+            headers=headers,
+            timeout=self._validate_timeout(timeout),
```
jordanambra, line 118: Validation should happen prior to usage. Can this be moved up? Also, i don't think it needs to return the timeout, just run the validate routine and raise an error.

jordanambra (changes requested):

jordanambra (approved):

PR #27, opened 2021-02-10
Feature/dc 960 change logic of from raw event 

oleksii-symon-corva-ai:
[DC-960 Change logic of  from_raw_event in each event type](https://corvaqa.atlassian.net/browse/DC-960)

- fixed loading logic in ScheduledEvent, StreamEvent and TaskEvent to load from objects instead of strings
- deleted `settings` fixture from conftest


corva/models/stream.py
```diff
@@ -129,24 +130,22 @@ def asset_id(self) -> Optional[int]:
...
+        )  # deepcopy an event as app_key field will be added to each dict
 
-        event_dicts = parse(List[dict], event)  # type: List[dict]
-
-        for event_dict in event_dicts:
+        for event_dict in event:
             if 'app_key' in event_dict:
                 raise ValueError(
-                    "app_key can't be set manually, it is extracted from env and set automatically."
+                    'app_key can\'t be set manually, it is extracted from env and set automatically.'
```
jordanambra, line 143: OK my dude, we don't have to be allergic to double quotes 😄

corva/models/stream.py
```diff
@@ -129,24 +130,22 @@ def asset_id(self) -> Optional[int]:
...
 
     @staticmethod
-    def from_raw_event(event: Union[str, List], **kwargs) -> List[StreamEvent]:
+    def from_raw_event(event: List[dict], **kwargs) -> List[StreamEvent]:
         app_key = kwargs['app_key']
 
-        parse = pydantic.parse_obj_as
-        if isinstance(event, str):
-            parse = pydantic.parse_raw_as
+        event = copy.deepcopy(
```
jordanambra, line 136: What's the reason we need to deep copy? Some of these events are pretty big and the memory profile of storing this will double.

tests/docs_src/test_tutorial_1.py
```diff
@@ -1,13 +1,20 @@
...
 from docs_src import tutorial_1_hello_world
 
 
-def test_tutorial(settings):
-    event = (
-        '[{"records": [{"asset_id": 0, "timestamp": 0}], '
-        '"metadata": {"app_stream_id": 0, "apps": {"%s": {"app_connection_id": 0}}}}]'
-    ) % settings.APP_KEY
+def test_tutorial():
+    event = [
```
jordanambra, line 8: Nice, this is much more readable

jordanambra (changes requested):
Just some minor questions and comments, other than that, this looks great.

corva/models/stream.py
oleksii-symon-corva-ai: it changed smh:) anyways after adding black to ci everything will be double quotes as it is a default

corva/models/stream.py
oleksii-symon-corva-ai: > What's the reason we need to deep copy?

StreamEvent.from_raw_event gets event of type List[dict]. Inside the function a new field `app_key` is added to each dict. If we don't copy each dict the caller of from_raw_event will get modified dicts without event expecting it.

> Some of these events are pretty big and the memory profile of storing this will double.

That's right, reworked the logic to copy only one dict at a time

ville (approved):

oleksii-symon-corva-ai:
merging, as all fixes are done. if there will be other comments, they will be applied in a patch

PR #29, opened 2021-02-23
Fix/setup py

oleksii-symon-corva-ai:


jordanambra:
Is there a task associated with this? Do we actually require these version upgrades for our customers or can they still use older versions of them?

oleksii-symon-corva-ai:
there is no task associated, I just saw that 1. versions in setup.py and requirements were different 2. packages option could be simplified

oleksii-symon-corva-ai:
> can they still use older versions of them

I think that minimum version in `install_requires` should be 1. >= than version in requirements and 2. < than next major library version (example https://github.com/tiangolo/fastapi/blob/33be5fc8baa02bae949e1986f1821dae8f84f487/pyproject.toml#L36)

setup.py
```diff
@@ -32,12 +33,7 @@
...
     keywords='corva, sdk',
-    packages=setuptools.find_packages(exclude=[".*", "data", "test"]),
-    install_requires=[
-        "pydantic >= 1.7.2",
-        "redis >= 3.5.3",
-        "requests >= 2.25.0",
-        "urllib3 >= 1.26.2"
-    ],
+    packages=["corva"],
+    install_requires=["pydantic >= 1.7.3", "redis >= 3.5.3", "requests >= 2.25.0"],
```
finomayato, line 37: requirement ranges should be closed to not to be updated to the major version e.g. `~=`/`>=,<=`

jordanambra, line 37: I thought this was easier to read with each requirement on its own line as well

setup.py
```diff
@@ -32,12 +33,7 @@
...
     license='The Unlicense',
```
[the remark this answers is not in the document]
jordanambra: Let me ask for support from the product team on this and we will follow up with a change, if necessary.

jordanambra (changes requested):

setup.py
oleksii-symon-corva-ai: added ranges

oleksii-symon-corva-ai: fixed

jordanambra (approved):

PR #30, opened 2021-02-24
Provide tools to conveniently test Corva apps

oleksii-symon-corva-ai:
[DC-925 Provide the best way for users to test their apps](https://corvaqa.atlassian.net/browse/DC-925)

The goal is to let users seamlessly test apps, without patching and even thinking about corva-sdk

What was done?

- Created a pytest plugin that automatically does the following:
  - sets up test environment vars used by the sdk 
  - does essential patches, that allow users write tests without thinking about sdk's internals.
- Added `corva_context` fixture to plugin, which imitates AWS lambda context and can be passed by users to their apps. It is needed as users don't know, what sdk expects to receive in context.
- Made all environment variables NOT optional, as it had to be done from the beginning.
 - Changed `Corva` class to search for `api key` only in context
- Separated `test_requirements.txt` and `requirements.txt`.

How the tests might look now? [link to gamma-depth PR](corva-ai/dev-center-gamma-depth-python#8)

```python3
# lambda_function.py
from corva import Corva


def user_stream_app(event, api, cache):
    ...


def lambda_handler(event, context):
    corva = Corva(context)
    corva.stream(user_stream_app, event)


# tests/test_app.py
from lambda_function import lambda_handler


# get the corva_context provided by plugin (no imports needed)
def test_app(corva_context):
    event = ...

    lambda_handler(event, corva_context)

    assert True
```


corva/testing.py
```diff
@@ -0,0 +1,78 @@
+import contextlib
+import functools
+import os
+from unittest import mock
+
+import fakeredis
+import pytest
```
oleksii-symon-corva-ai, line 7: should this be added to setup.py `install_requires`?

corva/testing.py
```diff
@@ -0,0 +1,78 @@
...
+
+from corva.configuration import Settings
+
+
+@pytest.fixture(scope='function')
```
oleksii-symon-corva-ai, line 12: maybe it'd be better to set `autouse=True`, so that users even dont know about the fixture

finomayato:
Looks very good! Can you make a corresponding PR in the user app, so we can see how it will work from the user perspective?

oleksii-symon-corva-ai:
@finomayato please, see user app changes https://github.com/corva-ai/dev-center-gamma-depth-python/pull/8

ville (approved):

finomayato (approved):

PR #31, opened 2021-03-02
Feature/dc 1135 fix infrastructure

oleksii-symon-corva-ai:
[DC-1135 Fix corva-sdk infrastructure](https://corvaqa.atlassian.net/browse/DC-1135)

- moved package files to src/ dir. [reason](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure)
- moved requirements to setup.py
- updated workflows/main.py


setup.py
```diff
@@ -27,23 +21,32 @@
...
+    packages=setuptools.find_packages("src"),
+    package_dir={"": "src"},
     install_requires=[
         "fakeredis >=1.4.5, <2.0.0",
         "pydantic >=1.7.3, <2.0.0",
         "redis >=3.5.3, <4.0.0",
         "requests >=2.25.0, <3.0.0",
         "requests-mock >=1.8.0, <2.0.0",
     ],
+    extras_require={
```
jordanambra, line 40: If you want to move away from having a version locked requirements.txt and putting everything into setup.py, you're going to need to start testing this with a Tox matrix. Otherwise, we have no real guaranteed working version that passes tests.

It's always best for development purposes to have packages locked to a specific version, but here with install_requires we want to allow flexibility on customer-facing installs only. It's OK that these concepts are separate.

jordanambra (changes requested):

setup.py
oleksii-symon-corva-ai: Default python projects need different requirements files, package python projects only need setup.py.

Only setup.py is needed because:

1. We run tests against the package as it will be installed by users, not against whatever the situation in project directory is. Users will never run requirements.txt only setup.py file.
2. If our tests fail, because of some dependency - we are lucky. We found and fixed the error before user encountered it.

Having only setup.py (or pyproject.toml for other packagers) is a usual thing for libraries to do:

- [fastapi](https://github.com/tiangolo/fastapi/blob/master/pyproject.toml)

- [pytest](https://github.com/pytest-dev/pytest)

About tox matrix. We can add that for test completeness in the future.

setup.py
```diff
@@ -27,23 +21,32 @@
     author='Jordan Ambra',
     author_email="jordan.ambra@corva.ai",
     url='https://github.com/corva-ai/python-sdk',
-    version=version,
-    classifiers=classifiers,
+    version="0.0.12",
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: returned the file, it will be shipped in the package for users to see

oleksii-symon-corva-ai:
Ping:) Can this be merged yet?

setup.py
```diff
@@ -27,23 +25,32 @@
...
-    ),
+    py_modules=[file.stem for file in pathlib.Path('src').glob('*.py')],
+    packages=setuptools.find_packages("src"),
+    package_dir={"": "src"},
     install_requires=[
         "fakeredis >=1.4.5, <2.0.0",
         "pydantic >=1.7.3, <2.0.0",
         "redis >=3.5.3, <4.0.0",
         "requests >=2.25.0, <3.0.0",
         "requests-mock >=1.8.0, <2.0.0",
```
finomayato, line 42: There is ~= notation you can use

finomayato (approved):
Looks good

setup.py
oleksii-symon-corva-ai: thanks

oleksii-symon-corva-ai:
Merging, because I'd like to move on with other PRs, that depend on this one.

@jordanambra we can continue the discussion if needed

PR #32, opened 2021-03-05
Scheduled app fixes and improvements

oleksii-symon-corva-ai:
This PR contains changes, which were discovered as essential, while developing scheduled version of gamma-depth [dev-center-gamma-depth-python/pull/9](https://github.com/corva-ai/dev-center-gamma-depth-python/pull/9).

Changes:

1. Added Api.get_dataset method
2. Changed `ScheduledEvent.schedule_end` and `ScheduledEvent.schedule_start` type to int
3. Added validator for `ScheduledEvent.schedule_end` and `ScheduledEvent.schedule_start`, which casts timestamp from millisecond to seconds, if needed
4. Made `ScheduledEvent.schedule_end` field optional
5. Added `plugin.patch_scheduled.patch_runner` to fix and simplify scheduled app testing
6. Fixed `setup.py` `install_requires` version ranges, as `~=` is not supported yet [link](http://peak.telecommunity.com/DevCenter/setuptools#declaring-dependencies)


src/corva/api.py
```diff
@@ -126,3 +128,59 @@ def _validate_timeout(self, timeout: int) -> bool:
...
+        Raises:
+          requests.HTTPError: if some request was unsuccessful.
+
+        Returns:
+          all data from dataset.
+        """
+
+        result = []
+
+        for skip in itertools.count(0, limit):
```
ville, line 166: We should not do this to prevent users from killing API/Mongo. For example, WITS collection has millions of records per asset. Just make user to pass the limit, max is 10 000. User can make the loop if more is needed.

ville (changes requested):

src/corva/api.py
oleksii-symon-corva-ai: What do you think about using generator, so that users can loop over data easier, like following:

```python3
def get_dataset(
    self,
    provider: str,
    dataset: str,
    query: dict,
    sort: dict,
    limit: int,
    skip: int = 0,
    fields: Optional[str] = None,
) -> Iterator[List[dict]]:

    for loop_skip in itertools.count(skip, limit):
        response = self.get(
            f"/api/v1/data/{provider}/{dataset}/",
            params={
                "query": json.dumps(query),
                "sort": json.dumps(sort),
                "fields": fields,
                "limit": limit,
                "skip": loop_skip,
            },
        )
        response.raise_for_status()

        new_data = list(response.json())

        yield new_data

        if len(new_data) != limit:
            break
```

Usage example:

```python3
gen = api.get_dataset(
    provider="corva",
    dataset="dataset",
    query={"asset_id": 123},
    sort={"timestamp": 1},
    limit=1000,
)

# get a chunk of data
data = next(gen)

# get all data
all_data = list(gen)
```

ville: It's really, really rare occasion when you want to fetch more than 10 000 records.

oleksii-symon-corva-ai: Thanks, reworked the function

src/plugin.py
```diff
@@ -103,40 +122,72 @@ def test_stream(self: Corva, fn, event, *args, **kwargs):
...
+            # context.api is accessed two times in scheduled_runner:
+            #   1. To pass an Api instance to the user function.
+            #   2. To make a POST API call.
+            # Patch context.api to return Mock instance, when accessed the second time.
+            # This is needed for following reasons:
+            #   1. To avoid POSTing in user tests.
+            #   2. To avoid adding our POST mock to users Api instance.
+            type(context).api = mock.PropertyMock(
+                side_effect=[context.api, mock.Mock()]
+            )
```
finomayato, line 179: `type(context).api` looks like magic. Is there any other way to mock `api` property?

oleksii-symon-corva-ai: In this case, I don't think there is an alternative. https://stackoverflow.com/a/54441868

oleksii-symon-corva-ai:
> Mocking functionality in `src/plugin.py` looks very complex. Let's hope it won't break :)

True, maybe we will be able to somehow simplify it in the future

setup.py
```diff
@@ -35,11 +35,10 @@
     packages=setuptools.find_packages("src"),
     package_dir={"": "src"},
     install_requires=[
-        "fakeredis ~=1.4.5",
-        "pydantic ~=1.7.3",
-        "redis ~=3.5.3",
-        "requests ~=2.25.0",
-        "requests-mock ~=1.8.0",
+        "fakeredis >=1.4.5, <2.0.0",
+        "pydantic >=1.7.3, <2.0.0",
```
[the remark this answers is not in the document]
finomayato: got it, thanks

finomayato (approved):
Looks good

ville (approved):
LGTM

src/corva/models/scheduled.py
```diff
@@ -48,6 +54,23 @@ class ScheduledEvent(BaseEvent):
...
+
+        Casts Unix timestamp from millisecond to seconds, if provided timestamp is in
+          milliseconds.
+        """
+
+        try:
+            # the function will raise, if provided timestamp is in milliseconds,
+            # because the year will always be too big to fit into the datetime instance
+            datetime.utcfromtimestamp(v)
+        except ValueError:
```
jordanambra, line 69: There's no guarantee that the data is going to be an integer in milliseconds if it throws a ValueError, is there? I believe this will throw an OverflowError. It may also be better to just check if the timestamp is > the max possible timestamp.

src/corva/models/scheduled.py
```diff
@@ -48,6 +54,23 @@ class ScheduledEvent(BaseEvent):
...
+            v /= 1000  # cast value from milliseconds to seconds
```
jordanambra, line 70: I think you can cast the value to an int here only, right?

src/plugin.py
```diff
@@ -2,18 +2,28 @@
...
 def pytest_load_initial_conftests(args, early_config, parser):
-    """Sets test environment variables."""
+    """Sets test environment variables.
+
+    This hook is triggered before loading the packages being tested.
+    If packages contain global variables, that read from env,
+    they will be able to read env values set by this hook.
+
+    Why does this hook get triggered before loading the packages being tested?
+      Current file, that that contains the hook is registered as pytest plugin through
```
jordanambra, line 22:
```suggestion
      The current file that contains the hook is registered as a pytest plugin through
```

src/plugin.py
```diff
@@ -46,14 +56,14 @@ def corva_context(corva_patch):
 
 @contextlib.contextmanager
 def patch_redis_adapter():
-    """Allows testing Corva apps without running real redis server.
+    """Allows testing Corva apps without need to run real Redis server.
```
jordanambra, line 59:
```suggestion
    """Allows testing of Corva apps without running a real Redis server.
```

src/plugin.py
```diff
@@ -46,14 +56,14 @@ def corva_context(corva_patch):
...
 
-    Internally Corva uses Redis as cache. The fixture allows testing Corva apps
-    without running real redis server. It patches RedisAdapter to use fakeredis
-    instead of redis. fakeredis is a library that simulates talking to a real redis
-    server.
+    Internally Corva uses Redis as cache. This function patches RedisAdapter to use
+    fakeredis instead of redis. fakeredis is a library that simulates talking to a
+    real Redis server. This way the function allows testing Corva apps without running
+    real Redis server.
```
jordanambra, line 64:
```suggestion
    a real Redis server.
```

jordanambra (changes requested):
Looks pretty good, just a few minor questions

src/corva/models/scheduled.py
oleksii-symon-corva-ai: if value doesn't need cast from ms to sec and comes as float, it will be returned as float
so we make sure that return type is int, by casting it before returning

oleksii-symon-corva-ai: > I think you can cast the value to an int here only, right?

that is right. as our validation occurs after pydantic one and validator param `always=False`: we always get int as input. so if cast from ms to sec not needed, we can just return the value

src/corva/models/scheduled.py
oleksii-symon-corva-ai: Good idea, reworked the logic

oleksii-symon-corva-ai:
Did all the fixes, merging to proceed with other tasks

PR #34, opened 2021-03-16
Feature/dc 755 add task app

oleksii-symon-corva-ai:
[DC-755 Add task app type to Corva](https://corvaqa.atlassian.net/browse/DC-755)

- Added task app type to Corva
- simplified patching in plugin._patch_scheduled


src/corva/runners/task.py
```diff
@@ -0,0 +1,47 @@
...
+    api: Api, task_id: str, status: TaskStatus, data: dict
+) -> requests.Response:
+    """Updates the task. Should never raise."""
+
+    return api.put(path=f'v2/tasks/{task_id}/{status}', data=data)
+
+
+def task_runner(fn: Callable, context: TaskContext):
+    try:
+        # TODO: in case this raises, should exc be added to fail_reason?
```
oleksii-symon-corva-ai, line 26: @ville what do you think about this?

ville: If this throws an error, then most likely update_task_data will fail as well.

ville (approved):
LGTM

src/corva/runners/task.py
oleksii-symon-corva-ai: Yes. `update_task_data` doesnt raise, so the task will just silently finish without updating the task. So the task will stay in running state

PR #35, opened 2021-03-26
Feature/dc 1236 simplify app testing

oleksii-symon-corva-ai:
_Most of the changes have to do with tests and readme, so don't be afraid of changed number of lines._

[DC 1236 Simplify app testing](https://corvaqa.atlassian.net/browse/DC-1236)

- Split kafka events from events passed to apps. App events have flat structure and contain only essential data.
- Added `TestClient`, which contains `run` function, that should be used for app testing. 
- Added `app_runner` fixture to plugin for ease of access to `TestClient.run` function.
- Deleted all patches but `_patch_redis_adapter` from plugin.
- Split stream event into `StreamTimeEvent` and `StreamDepthEvent`.
- Made filtering automatic for stream events.

How do the tests look now and why they are easy to write?

- Old problem: users had to define and provide kafka events to test the apps. Kafka events have different types (List[List[dict]], List[dict], dict), dicts contain many unkown to users fields, dicts dont have a strict structure. Solution: user can now define and provide _app events_ to `app_runner` (not kafka ones). App events have flat well-defined structure, minimum number of fields and type hints. 
- Old problem: users had to provide custom AWS Lambda context to test the apps. Solution: `app_runner` provides context object automatically.
- Old problem: when testing app with single event, users would get returned result in list. Solution: `app_runner` always returns single result.

Task app test:

```python
from corva import Corva, TaskEvent


def task_app(event, api):
    return 'Hello, World!'


def lambda_handler(event, context):
    return Corva(context).task(fn=task_app, event=event)


def test_task_app(app_runner):
    event = TaskEvent(asset_id=0, company_id=0)

    result = app_runner(fn=lambda_handler, event=event)

    assert result == 'Hello, World!'

```

Scheduled app test:

```python
from corva import Corva, ScheduledEvent


def scheduled_app(event, api, cache):
    return 'Hello, World!'


def lambda_handler(event, context):
    return Corva(context).scheduled(fn=scheduled_app, event=event)


def test_scheduled_app(app_runner):
    event = ScheduledEvent(asset_id=0, time_from=0, time_to=0)

    result = app_runner(fn=lambda_handler, event=event)

    assert result == 'Hello, World!'
```

Stream app test:

```python
from corva import (
    Corva,
    StreamDepthEvent,
    StreamDepthRecord,
    StreamTimeEvent,
    StreamTimeRecord,
)


def stream_app(event, api, cache):
    return 'Hello, World!'


def lambda_handler(event, context):
    return Corva(context).stream(fn=stream_app, event=event)


def test_stream_time_app(app_runner):
    event = StreamTimeEvent(
        asset_id=0, company_id=0, records=[StreamTimeRecord(timestamp=0)]
    )

    result = app_runner(fn=lambda_handler, event=event)

    assert result == 'Hello, World!'


def test_stream_depth_app(app_runner):
    event = StreamDepthEvent(
        asset_id=0, company_id=0, records=[StreamDepthRecord(measured_depth=0)]
    )

    result = app_runner(fn=lambda_handler, event=event)

    assert result == 'Hello, World!'
```


src/corva/models/base.py
```diff
@@ -5,64 +5,62 @@
...
+class EventConfig:
+    extra = pydantic.Extra.ignore
+    allow_mutation = False
 
 
-class CorvaBaseModel(pydantic.BaseModel):
-    Config = CorvaModelConfig
+class CorvaBaseEvent(pydantic.BaseModel):
+    class Config(EventConfig):
+        pass
```
finomayato, line 21: I'd rather choose the explicit version of it. Inherited empty class can be confusing

finomayato (approved):
I like the change overall 👍

src/corva/models/base.py
```diff
@@ -5,64 +5,62 @@
...
 
-    event: BaseEventTV
+    event: RawBaseEventTV
     settings: Settings
     api: Api
-    _cache: Optional[RedisState] = None
-
-    user_result: Any = None
-
+    cache_: Optional[RedisState] = None
```
jordanambra, line 50: Why was this renamed to an underscore with a suffix?

src/corva/runners/task.py
```diff
@@ -1,12 +1,18 @@
...
 import requests
 
 from corva.api import Api
-from corva.models.task import TaskContext, TaskEvent, TaskStatus
+from corva.models.task import TaskContext, TaskEvent
 
 
-def get_task_data(api: Api, task_id: str) -> TaskEvent:
+class TaskStatus(enum.Enum):
+    fail = 'fail'
```
jordanambra, line 11: I don't know if this matters, but the actual task status is `failed` or `succeeded`. The endpoints end with /fail or /success.

src/corva/utils.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any, Optional
```
jordanambra, line 1: Can we put these functions into an appropriate class or module? A utils.py file usually ends up just being spaghetti after a while.

jordanambra (changes requested):

src/corva/models/base.py
oleksii-symon-corva-ai: We need `EventConfig` to be shared by `CorvaBaseEvent` and `CorvaBaseGenericEvent`.

We have a few options for this.

```python
import pydantic
from typing import ClassVar


class EventConfig:
    ...


# Option 1
class CorvaBaseEvent1(pydantic.BaseModel):
    Config = EventConfig


# Option 2
class CorvaBaseEvent2(pydantic.BaseModel):
    Config: ClassVar = EventConfig  # type: ignore


# Option 3
class CorvaBaseEvent3(pydantic.BaseModel):
    class Config(EventConfig):
        pass
```

If we use `Option 1` pycharm would list Config as a parameter to model, which may be distracting.
image.

`Option 2` and `Option 3` dont have this issue.

Which one is better?

src/corva/models/base.py
oleksii-symon-corva-ai: Pydantic does some magic with fields prefixed with `_` like `_cache`. For some reason tests started to fail for old name.

src/corva/runners/task.py
oleksii-symon-corva-ai: Thanks, we don't parse actual task statuses, only set new ones using `v2/tasks/{task_id}/fail` and `v2/tasks/{task_id}/success`

src/corva/models/base.py
jordanambra: I liked it how you wrote it, actually. It solves the problem without extra type overhead.

jordanambra (approved):

src/corva/models/scheduled.py
```diff
@@ -67,16 +52,22 @@ def set_timestamp(cls, v: int) -> int:
 
         return v
 
+    @pydantic.root_validator(pre=False, skip_on_failure=True)
+    def set_time_from(cls, values: dict) -> dict:
+        """Calculates time_from field."""
+
+        values["time_from"] = int(values["schedule_start"] - values["interval"] + 1)
```
ville, line 59: Why `+1`?

oleksii-symon-corva-ai: Because `time_from` means _the first not processed time._ https://github.com/corva-ai/python-sdk/blob/94b2d697609db9a77e1a54d9c51efbe51d2874dd/src/corva/models/scheduled.py#L16

ville: Does the users understand this? This must be synced with Node SDK.

oleksii-symon-corva-ai: This field will be easier for users to understand. Compare following exactly the same examples:

Old SDK stream app usage. Pay attention to contents of `query["timestamp"]`.

```python3
import pydantic


class StreamEvent(pydantic.BaseModel):
    asset_id: int
    schedule_start: int
    interval: int


def stream_app(event: StreamEvent, api, cache):
    raw_records = api.get_dataset(
        provider="",
        dataset="",
        query={
            "asset_id": event.asset_id,
            "timestamp": {
                "$gt": event.schedule_start - event.interval,
                "$lte": event.schedule_start,
            },
        },
        sort={},
        limit=1000,
    )
```

New SDK stream app usage. Pay attention to contents of `query["timestamp"]`.

```python3
import pydantic


class StreamEvent(pydantic.BaseModel):
    asset_id: int
    time_from: int
    time_to: int


def stream_app(event: StreamEvent, api, cache):
    raw_records = api.get_dataset(
        provider="",
        dataset="",
        query={
            "asset_id": event.asset_id,
            "timestamp": {
                "$gte": event.time_from,
                "$lte": event.time_to,
            },
        },
        sort={},
        limit=1000,
    )
```

oleksii-symon-corva-ai: I've renamed the fields and expanded ScheduledEvent docstr [here](https://github.com/corva-ai/python-sdk/pull/35/commits/d9d89486900859544df9e2a8e2deaa3fd79a3c60)

PR #36, opened 2021-04-08
Feature/implement logging

oleksii-symon-corva-ai:
Rationale.
Users need a safe way to do logging in apps. Safe means - apps should not be able to log too much, as this will incur huge expenses for users.

[DC-1039 Implement logging thresholds](https://corvaqa.atlassian.net/browse/DC-1039)
[DC-1073 Prefix log messages with Asset ID and App Connection ID](https://corvaqa.atlassian.net/browse/DC-1073)

- Added logger to SDK
- Added `LOG_THRESHOLD_MESSAGE_SIZE` and `LOG_THRESHOLD_MESSAGE_COUNT` env variables, that limit logging
- Deleted logger usage from state/ dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,16 +269,18 @@ other operations with data.
 ## Logging
 As apps are executed very frequently
 (once a second or so),
-unlimited logging can lead to unexpected huge bills.
+unlimited logging can lead to huge amounts of data.
 
 The SDK provides a `Logger` object,
 which is a safe way for logging in apps.
 
 The `Logger` is a `logging.Logger` instance
-and so should be used like every Python logger.
+and should be used like every other Python logger.
 
-The `Logger` protects from huge bills
-by having following features:
+The `Logger` has following features:
+1. Log messages are injected with contextual information,
+   which makes it easy to filter through logs
+   while debugging issues.
 1. Log message length is limited. 
    Too long messages are truncated to not exceed the limit. 
    Set by `LOG_THRESHOLD_MESSAGE_SIZE` env variable. 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ The `Logger` has following features:
    Set by `LOG_THRESHOLD_MESSAGE_COUNT` env variable.
    Default value is `15`.
 3. Logging level can be set using `LOG_LEVEL` env variable.
-   Default value is `WARN`.
+   Default value is `INFO`.
 
 #### Logger usage example
 
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ Corva python-sdk is a framework for building
 - [Event](#event)
 - [Api](#api)
 - [Cache](#cache)
+- [Logging](#logging)  
 - [Testing](#testing)  
 - [Contributing](#contributing)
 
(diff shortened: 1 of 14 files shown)


ville (commented):
There should be 2 different limits. Limit for max message size and limit for logged message count.

oleksii-symon-corva-ai:
> There should be 2 different limits. Limit for max message size and limit for logged message count.

It'd better to have one limit - `max_total_number_of_output_symbols`, which is essentially equivalent to `max_message_size * logged_message_count`.

Why so?

1.  The costs are the same in both cases as total number of allowed output symbols is the same.
2. User gets more flexibility for logging. Sometimes user may need to log a long stack trace, which can be bigger than `max_message_size` - in this case the information is lost. With current implementation, there is smaller chance of such thing happening. E.g. usual logging messages are on average not to big, maybe, up to 90 symbols. Imagine following scenario `max_total_number_of_output_symbols=15000`, user starts logging: 1st log 50 symbols -> 2nd log 50 symbols -> 3rd log exception - number of symbols left 14900, which should be enough to not lose the log.

src/corva/logger.py
```diff
@@ -1,58 +1,117 @@
...
+    corva_handler = CorvaLoggerHandler(
+        max_chars=SETTINGS.LOG_MAX_CHARS, logger=DEFAULT_LOGGER
+    )
+
+    corva_handler.setLevel(SETTINGS.LOG_LEVEL)
+
+    # add formatter
+    corva_formatter = logging.Formatter(
+        '%(asctime)s.%(msecs)03dZ %(aws_request_id)s %(levelname)s '
+        '%(asset_id)s %(app_connection_id)s | %(message)s\n',
```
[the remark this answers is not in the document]
oleksii-symon-corva-ai: Yes, it does.

ville:
> > There should be 2 different limits. Limit for max message size and limit for logged message count.
> 
> It'd better to have one limit - `max_total_number_of_output_symbols`, which is essentially equivalent to `max_message_size * logged_message_count`.
> 
> Why so?
> 

This why:

```
2021-04-08T14:07:02.656Z very long message very long message very long message very long message very long message...
2021-04-08T14:07:03.656Z Disabling the logging as maximum number of logged characters was reached
```

vs

```
2021-04-08T14:07:02.656Z very long message...
2021-04-08T14:07:03.656Z some other message
2021-04-08T14:07:04.656Z some other message
2021-04-08T14:07:05.656Z some other message
2021-04-08T14:07:06.656Z Disabling the logging as maximum number of log messages reached
```

So if they for example log some object and that object is very large, no other logs will be written

oleksii-symon-corva-ai:
Users should be responsible for what they log (e.g. responsible for logging a long object and using all capacity, as they can control this); in case with exception users are not really in charge.

Looks like there is a trade-off and no perfect solution.

src/corva/logger.py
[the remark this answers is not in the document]
ville: Let's get rid of that by adding that space to `f' AC={app_connection_id}'`

ville:
> Users should be responsible for what they log

In perfect world, yes. But if even we can't achieve that, there's no way Dev Center users can. I checked Rollbar and Python stacktraces there seems to be less than 1000 characters.

ville:
What about adding a parameter to override that default limit per message?

oleksii-symon-corva-ai:
> What about adding a parameter to override that default limit per message?

We can do that, it'll require a descent amount of work though

What if we proceed with current changes and add new functionality as required? Because currently we are just guessing and can overcomplicate

ville:
> > What about adding a parameter to override that default limit per message?
> 
> We can do that, it'll require a descent amount of work though
> 
> What if we proceed with current changes and add new functionality as required? Because currently we are just guessing and can overcomplicate

Node SDK has already those 2 limits, we want to keep the functionality same.

oleksii-symon-corva-ai:
> Node SDK has already those 2 limits, we want to keep the functionality same.

Understandable, having `max_message_size` limit can be reasonable. Let's get some other opinions about our conversation @jordanambra @finomayato, what do you think?

As for `logged_message_count` - it is worse, than having `max_total_number_of_output_symbols`, because users won't be logging 1000 chars in every message, that way never reaching the logging limit, while also being constained to 15 messages.

ville:
> As for `logged_message_count` - it is worse, than having `max_total_number_of_output_symbols`, because users won't be logging 1000 chars in every message, that way never reaching the logging limit, while also being constained to 15 messages.

They serve different purpose, this would create 15 000 log events:
```
for record in very_large_list:
  logger.info(record.version)
```

finomayato (approved):
Both solutions discussed in this PR are equal to me. Both are an educated guess how the user works with logging. I propose to move with the easiest one

jordanambra:
We should match what the Node SDK does. We should use the same variable names as the Node SDK too.

oleksii-symon-corva-ai:
> We should match what the Node SDK does. We should use the same variable names as the Node SDK too.

@jordanambra 
SDKs should match for sure. But which solution will be more valuable and easier for users? It's hard to imagine an app developer scanning the app and counting number of log messages to not exceed the limit.

jordanambra:
I think the way that Sentry and Rollbar handle this is good. By default there is a fairly small message size that you can override if needed through settings.

README.md
```diff
@@ -265,6 +266,54 @@ other operations with data.
...
+1. Log message length is limited. 
+   Too long messages are truncated to not exceed the limit. 
+   Set by `LOG_THRESHOLD_MESSAGE_SIZE` env variable. 
+   Default value is `1000` symbols or bytes.
+2. Number of log messages is limited.
+   After reaching the limit logging gets disabled.
+   Set by `LOG_THRESHOLD_MESSAGE_COUNT` env variable.
+   Default value is `15`.
+3. Logging level can be set using `LOG_LEVEL` env variable.
+   Default value is `WARN`.
```
ville, line 293: Should default level be INFO? What do you think?

ville: Node SDK uses lowercase characters to set log level. Is it possible to make these the same? @katsanva

katsanva: Either node or python needs to implement some changes/mapping so that we will be on the same page.
I think this should be done in `python-sdk` because we have some amount of apps written and configured in node already
The supported list of log levels are `'fatal', 'error', 'warn', 'info', 'debug'`

oleksii-symon-corva-ai: This is where python and nodejs differ. e.g there is no `fatal` in python. I think we should leave log levels as expected by users of corresponding languages, so there is less learning

katsanva: @oleksii-symon-corva-ai disagree. I think we need to provide interchangeable abilities in SDK so that we can apply the language-agnostic configuration to the apps

ville: Users define the these env variables using manifest.json file.

katsanva: IMO it's much easier to explain to the user that manifest.json can contain
```
LOG_LEVEL is used to configure log level, possible values are 'fatal', 'error', 'warn', 'info', 'debug'
``` 

vs 

```
LOG_LEVEL is used to configure log level. 
If you are using nodejs possible values are 'fatal', 'error', 'warn', 'info', 'debug'. 
If you are going to use Python please use 'DEBUG', 'INFO', 'WARN', 'ERROR'
```

ville:
I'm fine merging this as is but the log level thing must be sorted out

ville (approved):

PR #47, opened 2021-07-08
Fix/dc 1806 delete obsolete logging

oleksii-symon-corva-ai:
### Rationale
There is an issue of the same exception being logged two times in stream and scheduled apps.

### Changes
Moved logging of exceptions out of `LoggingContext.__exit__` method to have better control.

[JIRA ticket](https://corvaqa.atlassian.net/browse/DC-1806)

#### TODO
- [ ] Update CHANGELOG.md

### DIFF
--- a/src/corva/handlers.py
+++ b/src/corva/handlers.py
@@ -26,26 +26,31 @@ def base_handler(raw_event_type: Type[RawBaseEvent]) -> Callable:
                 handler=logging.StreamHandler(stream=sys.stdout),
                 logger=CORVA_LOGGER,
             ) as logging_ctx:
-                context = CorvaContext.from_aws(
-                    aws_event=aws_event, aws_context=aws_context
-                )
+                try:
+                    context = CorvaContext.from_aws(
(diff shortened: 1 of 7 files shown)


src/corva/handlers.py
```diff
@@ -26,26 +26,31 @@ def wrapper(aws_event: Any, aws_context: Any) -> List[Any]:
...
 
-                results = [
-                    func(raw_event, api, context.aws_request_id, logging_ctx)
-                    for raw_event in raw_events
-                ]
+                    return results
 
-                return results
+                except Exception:
+                    CORVA_LOGGER.exception('The app failed to execute.')
```
oleksii-symon-corva-ai, line 52: Should we do this logging? Lambda will log unhandled exception itself, but without extra info like asset id and app connection id

jordanambra: I think having the additional logging is good. We are always looking at issues by asset.

jordanambra (approved):

finomayato (approved):
Looks good. Though, not sure if we want to test the exact messaging for exceptions

PR #51, opened 2021-07-30
Feature/dc 1919 3 split scheduled event

oleksii-symon-corva-ai:
### Rationale
There are 3 types of scheduled events, we need to support all of them.

### Changes
* Deprecated old `ScheduledEvent`
* Added new `ScheduledDataTimeEvent`, `ScheduledDepthEvent` and `ScheduledNaturalTimeEvent` 

[JIRA ticket](https://corvaqa.atlassian.net/browse/DC-1919)

#### TODO
- [x] Update CHANGELOG.md


src/corva/models/scheduled/scheduled.py
```diff
@@ -0,0 +1,95 @@
...
+              3 ft.  |     top depth
+                     |
+              6 ft.  |     bottom depth
+
+    Attributes:
+        asset_id: asset id.
+        company_id: company id.
+        top_depth: start depth in ft., covered by this event. Use exclusively.
+        bottom_depth: end depth in ft., covered by this event. Use inclusively.
+        log_identifier: TODO
```
oleksii-symon-corva-ai, line 68: @ville could you help with description, please

src/corva/models/scheduled/scheduled.py
```diff
@@ -0,0 +1,95 @@
...
+    top_depth: float
+    bottom_depth: float
+    log_identifier: str
+    interval: float
+
+
+class ScheduledNaturalEvent(ScheduledEvent):
+    """Natural scheduled event data.
+
+    TODO
```
oleksii-symon-corva-ai, line 83: @ville could you help with description, please

src/corva/models/scheduled/scheduled.py
ville: Something like this:
`log_identifier: app stream log identifier. Used to scope data by stream, asset might be connected to multiple depth based logs.`

src/corva/models/scheduled/scheduled.py
ville: **ScheduledTimeEvent**
> Time scheduled apps are run with time frequency set up by user

This should include mention that the frequency when schedules are triggered is based on the incoming data.


**ScheduledNaturalEvent**

Natural Time scheduled apps are run with time frequency set up by user based on the actual time instead of data time.

src/corva/models/scheduled/scheduled.py
```diff
@@ -0,0 +1,95 @@
+import pydantic
+
+from corva.models.base import CorvaBaseEvent
+
+
+class ScheduledEvent(CorvaBaseEvent):
+    """Base class for scheduled event data."""
+
+
+class ScheduledTimeEvent(ScheduledEvent):
```
ville, line 10: Should this be called `ScheduledDataTimeEvent`?

src/corva/models/scheduled/scheduled.py
```diff
@@ -0,0 +1,95 @@
...
+
+    asset_id: int
+    company_id: int
+    top_depth: float
+    bottom_depth: float
+    log_identifier: str
+    interval: float
+
+
+class ScheduledNaturalEvent(ScheduledEvent):
```
ville, line 80: Should this be called ScheduledNaturalTimeEvent?

setup.cfg
```diff
@@ -16,11 +16,15 @@ branch = True
...
 [coverage:report]
 precision = 2
-fail_under = 97.54
+fail_under = 98.79
 skip_covered = True
 show_missing = True
 exclude_lines =
     @abc.abstractmethod
+omit =
+    docs/src/logging/tutorial003.py
```
jordanambra, line 25: Maybe just totally omit the docs/src directory?

src/corva/models/scheduled/initial.py
```diff
@@ -0,0 +1,8 @@
+from corva.models.base import CorvaBaseEvent
+from corva.models.scheduled.scheduler_type import SchedulerType
+
+
+class InitialScheduledEvent(CorvaBaseEvent):
```
jordanambra, line 5: What's the purpose here, vs RawScheduledEvent?

src/corva/models/scheduled/raw.py
```diff
@@ -0,0 +1,122 @@
...
+        events = [
+            initial_event.scheduler_type.raw_event.parse_obj(sub_event)
+            for initial_event, sub_event in zip(initial_events, event)
+        ]
+
+        return events
+
+    def set_schedule_as_completed(self, api: Api) -> None:
+        """Sets schedule as completed."""
+        api.post(path=f'scheduler/{self.schedule_id}/completed')
```
jordanambra, line 68: Will this retry on failure?

jordanambra (changes requested):

src/corva/models/scheduled/scheduled.py
oleksii-symon-corva-ai: Thanks, now I see the difference

src/corva/models/scheduled/raw.py
oleksii-symon-corva-ai: No, there is no retries in Api class

src/corva/models/scheduled/initial.py
oleksii-symon-corva-ai: We don't know which subtype of `RawScheduledEvent` to use without `InitialScheduledEvent.scheduler_type` field

oleksii-symon-corva-ai: Actually, yes, we can delete InitialScheduledEvent

setup.cfg
oleksii-symon-corva-ai: docs/src should be included in coverage to make sure that all code snippets from docs are tested. those tutorials are ommited because they display usage of `sentry`, `rollbar` and `raygun`, which cannot be tested

src/corva/models/scheduled/raw.py
oleksii-symon-corva-ai: We should probably allow users to provide their own transport or something to easily enable retries

src/corva/__init__.py
```diff
@@ -15,7 +19,9 @@
     'Api',
     'Cache',
     'Logger',
-    'ScheduledEvent',
```
oleksii-symon-corva-ai, line 18: This will be not backward compatible, we will release version 2.0.0, is it okay @jordanambra @ville ?

ville: Yes, release it first as beta/rc for QA testing

ville (approved):
Looks good to me

src/corva/__init__.py
oleksii-symon-corva-ai: I mean we have 2 options:

1. 
    * Break backward compatibility by totally deleting `ScheduledEvent`
    * Let users know about breaking change by releasing next major version of the library - v2.0.0
 
2.
    * Save backward compatibility by doing following in \_\_init\_\_.py
```python
ScheduledEvent = ScheduledDataTimeEvent


def __getattr__(name):
    if name == "ScheduledEvent":
        warnings.warn(
            "corva.ScheduledEvent usage is deprecated and will be deleted in v2.0.0 use corva.ScheduledDataTimeEvent instead"
        )
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

Which one is preffered?

PR #56, opened 2021-12-30
DC-2617 support secrets

oleksii-symon-corva-ai:
### Rationale
Apps need a way to access secrets.

### Changes
Added secrets support.

[JIRA ticket](https://corvaqa.atlassian.net/browse/DC-2617)

#### TODO
- [X] Update CHANGELOG.md

### DIFF
--- a/src/corva/service/service.py
+++ b/src/corva/service/service.py
@@ -7,13 +7,11 @@ from corva.service.api_sdk import ApiSdkProtocol
 
 def run_app(
     has_secrets: bool,
-    app_connection_id: int,
+    app_id: int,
     api_sdk: ApiSdkProtocol,
     app: Callable[[], Any],
 ) -> Any:
-    secrets = (
-        api_sdk.get_secrets(app_connection_id=app_connection_id) if has_secrets else {}
-    )
+    secrets = api_sdk.get_secrets(app_id=app_id) if has_secrets else {}
 
     with mock.patch.dict(shared.SECRETS, values=secrets, clear=True):
         result = app()
--- a/src/corva/service/service.py
+++ b/src/corva/service/service.py
@@ -7,11 +7,11 @@ from corva.service.api_sdk import ApiSdkProtocol
 
 def run_app(
     has_secrets: bool,
-    app_id: int,
+    app_key: str,
     api_sdk: ApiSdkProtocol,
     app: Callable[[], Any],
 ) -> Any:
-    secrets = api_sdk.get_secrets(app_id=app_id) if has_secrets else {}
+    secrets = api_sdk.get_secrets(app_key=app_key) if has_secrets else {}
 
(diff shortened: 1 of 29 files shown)


oleksii-symon-corva-ai:
@ville @finomayato ready for review

ville:
I don't see this feature (we don't want to store this secret data to Redis):
```
Parameters should be cached for X minutes (5?)
- Name and value
- Cache needs to live in the global scope so that it gets shared between the invokes
```

src/corva/service/service.py
```diff
@@ -0,0 +1,19 @@
...
+from corva.service.api_sdk import ApiSdkProtocol
+
+
+def run_app(
+    has_secrets: bool,
+    app_key: str,
+    api_sdk: ApiSdkProtocol,
+    app: Callable[[], Any],
+) -> Any:
+    secrets = api_sdk.get_secrets(app_key=app_key) if has_secrets else {}
```
finomayato, line 14: why do we need a `get_secrets` param if we can try to get secrets right away and fill it with empty dict if none found?

src/corva/testing.py
```diff
@@ -50,3 +58,13 @@ def run(
...
                 ),
             )
+
+        if app is None:
+            return
+
+        return service.run_app(
+            has_secrets=True,
+            app_key='test_app_key',
+            api_sdk=FakeApiSdk(secrets={'test_app_key': secrets or {}}),
```
finomayato, line 68: It's possible to pass no `secrets` with `has_secrets=True` and wise versa. Seems, like one piece is redundant

src/corva/service/service.py
oleksii-symon-corva-ai: `get_secrets` is skipped if `has_secrets is False`.

`print(1) if False else print(2)` prints only 2

src/corva/testing.py
oleksii-symon-corva-ai: This is for app testing. If user passes secrets - he will see them in tests, otherwise empty dict is set.

I could have written, but this seems more complex
```python
  has_secrets = secrets is not None
  api_sdk_secrets = {'test_app_key': secrets} if has_secrets else None

  return service.run_app(
      has_secrets=has_secrets,
      app_key='test_app_key',
      api_sdk=FakeApiSdk(secrets=api_sdk_secrets),
      app=app,
  )
```

oleksii-symon-corva-ai:
@ville the caching was added https://github.com/corva-ai/python-sdk/pull/56/commits/9b578313a9d63e429a519a1fa99c44aa5a9223a8

src/corva/service/service.py
oleksii-symon-corva-ai: https://github.com/corva-ai/python-sdk/pull/56/commits/7c9413d47b00fdb2cc53c3b8b68b897cea08cac2

ville (approved):
Looks good to me

PR #69, opened 2022-10-17
Added the functionality to reuse cache in app_runner

mkamyab:
### Rationale
For integrated testing of scheduled or stream apps, we are relying on the cache to be reused. However, the `app_runner` does not provide handlers to support this functionality. I'm suggesting adding a `kwarg` in the `app_runner` to support this.

### Changes
- Added a `kwarg` to `app_runner` plugin
- Added a test case for this feature

#### TODO
- [ ] Update CHANGELOG.md

### DIFF
--- a/src/corva/api.py
+++ b/src/corva/api.py
@@ -33,24 +33,24 @@ class Api:
     @property
     def default_headers(self):
         return {
-            'Authorization': f'API {self.api_key}',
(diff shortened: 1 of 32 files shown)


mkamyab:
Too many changes due to `black`; will fix it and submit a new PR.

PR #82, opened 2023-11-27
New optional merge_events parameter for stream and scheduled apps

corva-alex-s:
### Rationale
We need to have ability to merge incoming events into one for "stream" and "scheduled" apps. For some use cases it'll greatly reduce number of IO operations.

This functionality(but only for scheduled apps) already exists in our node SDK, I'm mirroring it to python SDK. Functionality for "stream" type of apps is new, adding it for python SDK for now, in the near future we will do the same thing for node SDK(here https://corvaqa.atlassian.net/browse/DEVC-500)

### Changes
Added optional "merge_events" parameter for `stream` and `scheduled` decorators. If it's true - merge all events into one.

[[JIRA ticket]](https://corvaqa.atlassian.net/browse/DEVC-496)

#### TODO
- [ ] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
--  Added functionality to make retries for the failed HTTP requests.
 
 ### Added
-- Optional merge_events parameter to @stream and @scheduled decorators.
-Default is False, if True - merge all incoming events into one.
+- Optional merge_events parameter to @stream and @scheduled decorators. Default is False, if True - merge all incoming events into one.
+- Added functionality to make retries for the failed HTTP requests.
 
 
 # [1.10.0] - 2023-11-08


src/corva/handlers.py
```diff
@@ -544,3 +563,44 @@ def _get_custom_event_type_by_raw_aws_event(
...
+            ):
+                max_event_end = event[event_end]
+
+        aws_event[0][event_start] = min_event_start
+        if max_event_end:
+            aws_event[0][event_end] = max_event_end
+        aws_event = aws_event[0]
+        return aws_event
+
+    # stream event
```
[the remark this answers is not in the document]
tsilevych: most likely because task events can have random input parameters, depending on task functionality. How to merge them? On what criterion?

docs/modules/ROOT/examples/merging/tutorial001.py
```diff
@@ -0,0 +1,9 @@
+from corva import Api, StreamTimeEvent, stream
+
+
+# imagine we actually have 3 incoming events with 3 records each
+@stream(merge_events=True)
+def task_app(event: StreamTimeEvent, api: Api):
```
oleksii-symon-corva-ai, line 6:
```suggestion
def app(event: StreamTimeEvent, api: Api):
```

src/corva/handlers.py
```diff
@@ -544,3 +562,55 @@ def _get_custom_event_type_by_raw_aws_event(
...
+            aws_event[0]["records"].extend(event["records"])
+        aws_event = [aws_event[0]]
+        return aws_event
+
+    # unexpected event type, raise an exception
+    raise RuntimeError(
+        "merge_events parameter was passed to app type other than scheduled "
+        "or stream. Merge strategy is not implemented for "
+        f"{data_transformation_type} event types."
+    )
```
oleksii-symon-corva-ai, line 616: Wanted to share an interesting method on how to disallow passing other app types on Mypy static type check level. This will also probably make `test_events_not_merged_on_unexpected_event_type` obsolete
```python
if data_transformation_type is RawScheduledEvent:
    ...
elif data_transformation_type is RawStreamEvent:
    ...
else:
    typing.assert_never(data_transformation_type)
```

src/corva/handlers.py
```diff
@@ -544,3 +562,55 @@ def _get_custom_event_type_by_raw_aws_event(
...
+        )
+        min_event_start, max_event_end = (
+            aws_event[0][event_start],
+            aws_event[0].get(event_end),
+        )
+        for event in aws_event[1:]:
+            if event[event_start] < min_event_start:
+                min_event_start = event[event_start]
+            if (
+                max_event_end
```
oleksii-symon-corva-ai, line 592: If `aws_event[0].get(event_end)` returns `None`, `max_event_end` will never be reassigned

src/corva/handlers.py
```diff
@@ -544,3 +562,55 @@ def _get_custom_event_type_by_raw_aws_event(
...
+        # scheduled event
+        if not isinstance(aws_event[0], dict):
+            aws_event = list(itertools.chain(*aws_event))
+        is_depth = aws_event[0]["scheduler_type"] == SchedulerType.data_depth_milestone
+        event_start, event_end = (
+            ("top_depth", "bottom_depth")
+            if is_depth
+            else ("schedule_start", "schedule_end")
+        )
+        min_event_start, max_event_end = (
```
oleksii-symon-corva-ai, line 584:
```python
min_event_start = min(e[event_start] for e in aws_event)
max_event_start = max((e[event_end] for e in aws_event if event_end in e), default=None)
```

docs/modules/ROOT/examples/merging/tutorial001.py
oleksii-symon-corva-ai, line 1: This example needs to be run in test suite like so https://github.com/corva-ai/python-sdk/tree/master/tests/unit/test_docs

Using such code samples in docs allows us to show use cases to our users and also validate that they are working (by running tests on them). So, maybe consider adding `scheduled` app types here and maybe `test_merge_events_scheduled_event`, `test_merge_events_stream_event` will become obsolete too.

Or those tests may stay too as they have some specific logic with start/end times/depths

src/corva/handlers.py
```diff
@@ -544,3 +562,55 @@ def _get_custom_event_type_by_raw_aws_event(
...
+    Merges incoming aws_events into one.
+    Merge happens differently, depending on app type.
+    Only "scheduled" and "stream" type of apps can be processed here.
+    If somehow any other type is passed - raise an exception
+    """
+    if data_transformation_type == RawScheduledEvent:
+        # scheduled event
+        if not isinstance(aws_event[0], dict):
+            aws_event = list(itertools.chain(*aws_event))
+        is_depth = aws_event[0]["scheduler_type"] == SchedulerType.data_depth_milestone
```
oleksii-symon-corva-ai, line 578: can you compare str and enum using `==`? or should it be:
```python
aws_event[0]["scheduler_type"] == SchedulerType.data_depth_milestone.value
```

src/corva/handlers.py
oleksii-symon-corva-ai: Easier to just specify correct types in func definition `data_transformation_type: Union[Type[RawBaseEvent], ...]`

docs/modules/ROOT/examples/merging/tutorial001.py
```diff
@@ -0,0 +1,10 @@
+from corva import Api, StreamTimeEvent, stream, Cache
+
+
+# imagine we actually have 3 incoming events with 3 records each
+@stream(merge_events=True)
+def app(event: StreamTimeEvent, api: Api, cache: Cache):
+    # since we passed merge_events=True all 3 incoming events
+    # and their records will be merged into a single event with 9 records
+    assert len(event.records) != 9  # this will not fail
```
oleksii-symon-corva-ai, line 9: 3 events with 3 records each must equal to 9 records total?

docs/modules/ROOT/examples/merging/tutorial001.py
```diff
@@ -0,0 +1,11 @@
+from corva import Api  # <.>
```
oleksii-symon-corva-ai, line 1:
```suggestion
from corva import Api 
```

docs/modules/ROOT/examples/merging/tutorial002.py
```diff
@@ -0,0 +1,7 @@
+from corva import Api  # <.>
```
oleksii-symon-corva-ai, line 1:
```suggestion
from corva import Api
```

src/corva/handlers.py
oleksii-symon-corva-ai: `_merge_events` function receives aws event as dict, because parsing is done after it https://github.com/corva-ai/python-sdk/blob/DEVC-496-be-stream-spike-contact-devcenter-team-to-update-python-sdk-to-increase-data-batch-size/src/corva/handlers.py#L93

src/corva/handlers.py
```diff
@@ -544,3 +565,43 @@ def _get_custom_event_type_by_raw_aws_event(
...
+    elif data_transformation_type == RawStreamEvent:
+        # stream event
+        for event in aws_event[1:]:
+            aws_event[0]["records"].extend(event["records"])
+        aws_event = [aws_event[0]]
+        return aws_event
+
+    else:
+        # unexpected event type, raise an exception
+        assert_never(data_transformation_type)  # type: ignore
```
oleksii-symon-corva-ai, line 607: `typing_extensions` is not a direct dependency of the sdk, this may be a problem

what about just changing function header? If anyone passes other type - mypy will complain
```python
data_transformation_type: Union[Type[RawScheduledEvent], Type[RawStreamEvent]]
```

docs/modules/ROOT/examples/merging/tutorial001.py
oleksii-symon-corva-ai: assertion says `!=9` ?

oleksii-symon-corva-ai (approved):

PR #85, opened 2024-01-05
Most of the fields in "partial rerun merge" events should be optional

corva-alex-s:
### Rationale
During testing on QA we discovered that some of the fields might be missing or null, in such cases current SDK logic would not recognize them as "partial rerun merge" events and will fail to process them properly.

### Changes
changed most of the fields to be optional. The only fields required are those that we need to setup cache.

[JIRA ticket](https://corvaqa.atlassian.net/browse/DEVC-570)

#### TODO
- [x] Update CHANGELOG.md

### DIFF
--- a/tests/unit/test_partial_rerun_merge_app.py
+++ b/tests/unit/test_partial_rerun_merge_app.py
@@ -145,8 +145,8 @@ def test_event_parsing_not_failing_on_missing_field(context):
 
 def test_merge_events_does_not_fail_for_partial_rerun_merge_events(context, mocker):
     """
-    partial rerun merge events shouldn't fail to be processed when used in combination with
-    merge_events parameter
+    partial rerun merge events shouldn't fail to be processed when used in combination
+    with merge_events parameter
     """
 
     @corva.scheduled(merge_events=True)


src/corva/handlers.py
```diff
@@ -576,6 +576,11 @@ def _merge_events(
     Only "scheduled" and "stream" type of apps can be processed here.
     If somehow any other type is passed - raise an exception
     """
+    if not isinstance(
```
vladbagmet, line 579: Why should we run functionality to merge event if event does not have any data to merge?

src/corva/models/merge/merge.py
```diff
@@ -25,23 +27,23 @@ class PartialRerunMergeEvent(CorvaBaseEvent):
         run_until: run until
     """
 
-    event_type: EventType
-    partial_well_rerun_id: int
-    rerun_mode: RerunMode
-    start: int
-    end: int
+    event_type: Optional[EventType]
```
vladbagmet, line 30: This is a mandatory field, not optional. They have to send it. If they do not send it this is a problem that should be communicated.

src/corva/models/merge/merge.py
```diff
@@ -25,23 +27,23 @@ class PartialRerunMergeEvent(CorvaBaseEvent):
...
+    start: Optional[int]
+    end: Optional[int]
     asset_id: int
     rerun_asset_id: int
     app_stream_id: int
     rerun_app_stream_id: int
-    version: int
-    app_id: int
-    app_key: str
+    version: Optional[int]
```
vladbagmet, line 39: I think this field can not be optional as well. It's part of the mechanism of backwards-compatibility when the shape of data is changed.

tests/unit/test_partial_rerun_merge_app.py
```diff
@@ -121,3 +121,44 @@ def partial_rerun_merge_app(event, api, asset_cache, rerun_asset_cache):
...
+
+    @corva.scheduled(merge_events=True)
+    def scheduled_app(event, api, cache):
+        pytest.fail("Scheduled app was unexpectedly called!")
+
+    @corva.partial_rerun_merge
+    def partial_rerun_merge_app(event, api, asset_cache, rerun_asset_cache):
+        return True
+
+    event_with_missing_parameter: dict = deepcopy(RAW_EVENT)
```
vladbagmet, line 140: You can avoid using additional module by doing something like this
` event_with_missing_parameter = dict(RAW_EVENT)`

tests/unit/test_partial_rerun_merge_app.py
```diff
@@ -121,3 +121,44 @@ def partial_rerun_merge_app(event, api, asset_cache, rerun_asset_cache):
...
+    event_with_missing_parameter["data"]["end"] = None
+    partial_merge_app_response = scheduled_app(event_with_missing_parameter, context)[0]
+    assert partial_merge_app_response is True
+
+
+def test_merge_events_parameter_ignored_for_partial_rerun_merge(context):
+    """
+    merge_events parameter is ignored when incoming aws_event is not a list
```
vladbagmet, line 148: Good case to test. But where in your test do you actually test it?

vladbagmet (commented):
@corva-alex-s, I left some comments. The main question is why do we run functionality to merge event for event which does not have any data to merge?

src/corva/models/merge/merge.py
oleksii-symon-corva-ai, line 30: Ideally this field should be used to discriminate between different event types, so it should not be optional
https://docs.pydantic.dev/1.10/usage/types/#nested-discriminated-unions

oleksii-symon-corva-ai:
@vladbagmet are `HANDLERS` supposed to be empty?
https://github.com/corva-ai/python-sdk/blob/DEVC-570-be-sdk-fix-but-with-merge-events-parameter-for-slide-sheet-scheduler/src/corva/handlers.py#L42

tests/unit/test_partial_rerun_merge_app.py
```diff
@@ -121,3 +121,45 @@ def partial_rerun_merge_app(event, api, asset_cache, rerun_asset_cache):
...
+
+    @corva.partial_rerun_merge
+    def partial_rerun_merge_app(event, api, asset_cache, rerun_asset_cache):
+        return True
+
+    partial_merge_app_response = scheduled_app(RAW_EVENT, context)[0]
+    assert partial_merge_app_response is True
+
+
+def test_partial_rerun_decorator_can_be_called_directly(context):
```
oleksii-symon-corva-ai, line 164: What is the purpose of this test?

src/corva/models/merge/merge.py
oleksii-symon-corva-ai: I see, it is not optional below

vladbagmet:
> @vladbagmet are `HANDLERS` supposed to be empty? https://github.com/corva-ai/python-sdk/blob/DEVC-570-be-sdk-fix-but-with-merge-events-parameter-for-slide-sheet-scheduler/src/corva/handlers.py#L42

@oleksii-symon-corva-ai, handlers are initialised empty and then they are populated on the fly if we have any partial-merge decorator declared within an app 
https://github.com/corva-ai/python-sdk/blob/DEVC-570-be-sdk-fix-but-with-merge-events-parameter-for-slide-sheet-scheduler/src/corva/handlers.py#L551

tests/unit/test_partial_rerun_merge_app.py
vladbagmet: You are right, for nested dicts we should use deepcopy.

src/corva/handlers.py
vladbagmet: @corva-alex-s, partial merge rerun events handler was implemented before events merging functionality. What stopped you from learning how partial merge rerun functionality works before working on events merging functionality?

PR #86, opened 2024-02-05
DEVC-627 | Handle optional data field

vladbagmet:
### Rationale
Some old apps send events with empty data field which causes error like [this one](https://app.rollbar.com/a/corva/fix/item/Corva-Predictive-Drilling/61).

### Changes
Going forward, once event with empty data is received, no payload validation is raised.

[DEVC-627](https://corvaqa.atlassian.net/browse/DEVC-627)

#### TODO
- [x] Update CHANGELOG.md

[DEVC-627]: https://corvaqa.atlassian.net/browse/DEVC-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- For stream events data field can be None, not empty object.
+### Fixed
+- Filter out records with `None` data from stream time records.
 
 
 ## [1.11.2] - 2024-01-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- For stream events data field can be None, not empty object.
 
 
 ## [1.11.2] - 2024-01-05
--- a/src/corva/models/stream/raw.py
+++ b/src/corva/models/stream/raw.py
@@ -148,7 +148,7 @@ class RawStreamEvent(CorvaBaseEvent, RawBaseEvent):
     ) -> List[RawBaseRecord]:
         new_records = copy.deepcopy(self.records)
 
-        if self.is_completed and new_records:
+        if self.is_completed:
             new_records = new_records[:-1]  # remove "completed" record
 
         if old_max_record_value is None:
--- a/src/corva/models/stream/raw.py
+++ b/src/corva/models/stream/raw.py
@@ -19,7 +19,7 @@ class RawBaseRecord(CorvaBaseEvent, abc.ABC):
     company_id: int
     collection: str
 
-    data: Optional[dict] = {}
+    data: dict = {}
     metadata: dict = {}
 
     @property
@@ -73,7 +73,7 @@ if TYPE_CHECKING:
     RecordsDepth = Sequence[RawDepthRecord]
 else:
     RecordsBase = pydantic.conlist(RawBaseRecord, min_items=1)
-    RecordsTime = pydantic.conlist(RawTimeRecord, min_items=1)
+    RecordsTime = pydantic.conlist(RawTimeRecord, min_items=0)
     RecordsDepth = pydantic.conlist(RawDepthRecord, min_items=1)
 
 
@@ -105,7 +105,10 @@ class RawStreamEvent(CorvaBaseEvent, RawBaseEvent):
         There can only be 1 completed record always located at the end of the list.
         """
 
-        return self.records[-1].collection == 'wits.completed'
+        if not self.records:
+            return False
+
+        return self.records[-1].collection == "wits.completed"
 
     @property
     def max_record_value(self) -> Union[int, float]:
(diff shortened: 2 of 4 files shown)


skylax1 (approved):

src/corva/models/stream/stream.py
```diff
@@ -16,7 +16,7 @@ class StreamTimeRecord(CorvaBaseEvent):
     """
 
     timestamp: int
-    data: dict = {}
+    data: Optional[dict] = {}
```
oleksii-symon-corva-ai, line 19: User-facing events should not be changed, because this change is not backward-compatible

tests/unit/test_stream_app.py
```diff
@@ -615,3 +615,40 @@ def stream_app(event, api, cache):
...
+                    "timestamp": 1688999883,
+                    "version": 1,
+                }
+            ],
+        }
+    ]
+
+    result_event: StreamTimeEvent = stream_app(event, context)[0]
+
+    assert result_event is None
```
oleksii-symon-corva-ai, line 654: Interesting, how can result event be None?

src/corva/models/stream/stream.py
vladbagmet: This is addressed.

tests/unit/test_stream_app.py
vladbagmet: In case when we do not have any data to build an event, we probably should not run user's app at all. I've updated the test to reflect that.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- For stream events data field can be None, not empty object.
```
oleksii-symon-corva-ai, line 8:
```suggestion
### Fixed
- Filter out records with `None` data from stream time records.
```

oleksii-symon-corva-ai:
@ville We've added filtering to stream time events, do we need one for depth events too?

src/corva/models/stream/raw.py
```diff
@@ -145,7 +148,7 @@ def filter_records(
     ) -> List[RawBaseRecord]:
         new_records = copy.deepcopy(self.records)
 
-        if self.is_completed:
+        if self.is_completed and new_records:
```
oleksii-symon-corva-ai, line 151: Maybe we don't need this check as logic below will succeed for empty lists too?
```suggestion
        if self.is_completed:
```

vladbagmet: You are right, we do not need it. Code [is updated](https://github.com/corva-ai/python-sdk/pull/86/commits/136c4f37db617749eda8fe8d5a4a2f1bb0820308).

src/corva/models/stream/raw.py
```diff
@@ -165,33 +168,48 @@ def filter_records(
...
 
-        values["company_id"] = int(records[0].company_id)
+        if records:
+            values["company_id"] = int(records[0].company_id)
 
         return values
 
+    @pydantic.validator("records", pre=True)
+    def validate_records(cls, v):
+        if isinstance(v, List):
```
oleksii-symon-corva-ai, line 191: Is it a good idea long-term to use `typing.List` in `isinstance` checks? Simple `list` might be more robust, especially, if we migrate to 3.9+

src/corva/models/stream/raw.py
```diff
@@ -165,33 +168,48 @@ def filter_records(
...
+            return [
+                record
+                for record in v
+                if (
+                    (isinstance(record, dict) and record.get("data") is not None)
+                    or (hasattr(record, "data") and record.data is not None)
```
oleksii-symon-corva-ai, line 197: Can't imagine a use case for `hasattr` in this case, is this line really needed?

tests/unit/test_stream_app.py
```diff
@@ -615,3 +615,42 @@ def stream_app(event, api, cache):
...
+                    "provider": "corva",
+                    "timestamp": 1688999883,
+                    "version": 1,
+                }  # DEVC-627. This record should be filtered out because data is None.
+            ],
+        }
+    ]
+
+    _ = stream_app(event, context)[0]
+    assert True, "App call should be skipped"
```
oleksii-symon-corva-ai, line 656: This can never fail:)

src/corva/models/stream/raw.py
vladbagmet: But this change will break SDK for old Dev Center apps which are still on Python3.8 or am I missing anything?

oleksii-symon-corva-ai: It won't, e.g. below we use simple `dict` in `isinstance` check - same thing can be used here

src/corva/models/stream/raw.py
vladbagmet: Without it it will start to fail with AttributeError.
```
AttributeError: 'dict' object has no attribute 'data'
```

tests/unit/test_stream_app.py
vladbagmet: I agree. But if we remove it will hurt the readability because you expect to see asserts at the end of the test. But I can remove it if needed.

oleksii-symon-corva-ai (approved):
:fire: Let's see, what Ville says, whether we need this for depth events too and merge!

ville:
> @ville We've added filtering to stream time events, do we need one for depth events too?

No need.

